PR #14, opened 2022-10-12
fix: isolate i18next instance

StanleyDing:
### What this commit solves
If the project using this widget is also using i18next (like us), current code will initialize user's i18next twice. This PR creates a new instance that is provided through react provider, so it won't have side-effect on user's i18next instance.

### Things to improve in the future
Let users pass their own i18next instance or an option to overwrite current translation files.


chybisov:
@StanleyDing hi, thank you for this PR. Yes, this makes total sense. I'll make some tests, and we can merge it 🙂

PR #67, opened 2023-03-22
refactor: wallet management v2

Addminus:


packages/widget/src/providers/WalletProvider/WalletProvider.tsx
```diff
@@ -127,9 +147,17 @@ export const WalletProvider: FC<PropsWithChildren> = ({ children }) => {
       addChain,
       addToken,
       account,
-      provider,
+      provider: currentWallet?.account?.provider || undefined,
```
chybisov, line 150: `undefined` is unnecessary here.

chybisov (changes requested):

chybisov:
@Addminus 
image

Could you please fix these errors inside `widget-embedded/src/providers/WalletProvider.tsx` as well 🙂

packages/wallet-management/src/wallets.ts
```diff
@@ -1,124 +1,255 @@
...
-    icon: walletIcons.alphawallet,
-  },
-  atoken: {
-    name: 'AToken',
-    icon: walletIcons.atoken,
+const defaultWallet: Wallet = new InjectedConnector({
+  name: 'Your Default Wallet',
+  installed: () => !!(window as any).ethereum,
+  icon: walletIcons.metamask,
+});
```
chybisov, line 12: It looks like this is not used anywhere. Do we need it?

chybisov:
@Addminus let's continue here #69 😅

PR #86, opened 2023-05-03
feat: add HighValueLoss emitter event

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-2817


packages/widget/src/types/events.ts
```diff
@@ -5,13 +5,22 @@ export enum WidgetEvent {
   RouteExecutionUpdated = 'routeExecutionUpdated',
   RouteExecutionCompleted = 'routeExecutionCompleted',
   RouteExecutionFailed = 'routeExecutionFailed',
+  RouteHighValueLoss = 'highValueLoss',
```
chybisov, line 8: Let's name it the same everywhere, so it should be 

`RouteHighValueLoss = 'routeHighValueLoss',`

packages/widget/src/types/events.ts
```diff
@@ -5,13 +5,22 @@ export enum WidgetEvent {
...
+  toAmountUSD: string;
+  valueLoss: string;
 }
 
 export type WidgetEvents = {
   routeExecutionStarted: Route;
   routeExecutionUpdated: RouteExecutionUpdate;
   routeExecutionCompleted: Route;
   routeExecutionFailed: RouteExecutionUpdate;
+  highValueLoss: HighValueLossUpdate;
```
chybisov, line 23: And here it should be 

`routeHighValueLoss: RouteHighValueLossUpdate;`

packages/widget/src/types/events.ts
```diff
@@ -5,13 +5,22 @@ export enum WidgetEvent {
   RouteExecutionUpdated = 'routeExecutionUpdated',
   RouteExecutionCompleted = 'routeExecutionCompleted',
   RouteExecutionFailed = 'routeExecutionFailed',
+  RouteHighValueLoss = 'highValueLoss',
+}
+
+export interface HighValueLossUpdate {
```
chybisov, line 11: So here it should be 

`export interface RouteHighValueLossUpdate {`

packages/widget/src/pages/SwapPage/TokenValueBottomSheet.tsx
```diff
@@ -43,7 +44,19 @@ const TokenValueBottomSheetContent: React.FC<TokenValueBottomSheetProps> = ({
...
+    fromAmountUsd: route.fromAmountUSD,
+    gasCostUSD: route.gasCostUSD,
+    toAmountUSD: route.toAmountUSD,
+    valueLoss: Big(route.toAmountUSD || 0)
+      .div(Big(route.fromAmountUSD || 0).plus(Big(route.gasCostUSD || 0)))
+      .minus(1)
+      .mul(100)
+      .round(2, Big.roundUp)
+      .toString(),
+  });
```
chybisov, line 59: We can't just emit events in the middle of the render phase. This should be wrapped in `useEffect`. Also, later on, if we decide bottom sheet to stay mounted all the time this can cause issues of firing event even when we don't show the bottom sheet to the user. 

I think the best place to fire this event is after we call it to open: https://github.com/lifinance/widget/blob/5fc59bc97b899d5ced2a7fd2dc545470c8c0a976/packages/widget/src/pages/SwapPage/SwapPage.tsx#L63

Wdyt? 🙂

chybisov (changes requested):

packages/widget/src/pages/SwapPage/TokenValueBottomSheet.tsx
dennyscode: @chybisov Thanks for wrapping your head around this and finding that bug! 👍 
The approach you pointed out was a good idea on the issue which I was focussing on first: Triggering the Emit once the HighValueLoss-message pops up.
Now, as I continued to work on, I thought: The best would be to emit an event whenever a user actually continues: That will be useful for later tracking.

Having that message emit is not a bad thing, but what really matters is when the user continues.
What do you think of that approach? Sure, I will return to your solution if you prefer!

chybisov: Yes, `Continue` is also a great place to emit this event. I thought about this too, but I also thought maybe we want to track precisely the event of showing the bottom sheet 😅

packages/widget/src/pages/SwapPage/TokenValueBottomSheet.tsx
```diff
@@ -43,6 +42,7 @@ const TokenValueBottomSheetContent: React.FC<TokenValueBottomSheetProps> = ({
 }) => {
   const { t } = useTranslation();
   const ref = useRef<HTMLElement>();
+  const emitter = useWidgetEvents();
```
chybisov, line 45: Probably we don't need this anymore here.

packages/widget/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -50,6 +57,19 @@ export const SwapPage: React.FC = () => {
...
+          fromAmountUsd: route.fromAmountUSD,
+          gasCostUSD: route.gasCostUSD,
+          toAmountUSD: route.toAmountUSD,
+          valueLoss: Big(route.toAmountUSD || 0)
+            .div(Big(route.fromAmountUSD || 0).plus(Big(route.gasCostUSD || 0)))
+            .minus(1)
+            .mul(100)
+            .round(2, Big.roundUp)
+            .toString(),
+        });
```
chybisov, line 71: How about combining the creation of this object to some function and using both inside TokenValueLossBottomSheet component and here? 🙂

chybisov (changes requested):

packages/widget/src/pages/SwapPage/SwapPage.tsx
dennyscode: Good idea, and DRYer!

dennyscode: @chybisov I have moved that piece of code to re-use. It´s to be found in the SwapPage.tsx named **calcValueLoss**. Not sure if a better place would be the utils-folder?

chybisov: > @chybisov I have moved that piece of code to re-use. It´s to be found in the SwapPage.tsx named **calcValueLoss**. Not sure if a better place would be the utils-folder?

Yes, please, create a utils.ts inside the SwapPage folder.

chybisov (approved):

PR #87, opened 2023-05-03
fix: handle undefined current wallet

Addminus:


packages/wallet-management/src/walletAutomation.ts
```diff
@@ -1,67 +1,52 @@
...
-            addChain(provider, chainId).then((result) => resolve(result));
-          } else {
-            reject(error);
-          }
-        });
-      provider.once('chainChanged', (id: string) => {
-        if (parseInt(id) === chainId) {
-          resolve(true);
-        }
-      });
```
chybisov, line 32: The last time we talked about this, we decided to keep it. Could you please share why you removed it? 🙂

chybisov (changes requested):

packages/wallet-management/src/walletAutomation.ts
```diff
@@ -1,67 +1,52 @@
...
-      if (error.code !== MetaMaskProviderErrorCode.userRejectedRequest) {
-        addChain(provider, chainId).then((result) => resolve(result));
-      } else {
-        console.error(error);
-        resolve(false);
-      }
+  try {
+    await provider.send('wallet_switchEthereumChain', [
+      { chainId: getChainById(chainId).metamask?.chainId },
+    ]);
```
chybisov, line 17: Why we moved from `provider.request` to `provider.send`?

chybisov (changes requested):

packages/wallet-management/src/walletAutomation.ts
Addminus: The provider object changed its type. The new type does not have request. It has send. 
This change was necessary to facilitate wallet automations without knowing which wallet is currently being used. Not knowing which wallet is being used happens when an external walletmanagement has been set but no wallet automation functions.

packages/wallet-management/src/walletAutomation.ts
Addminus: The .send() works differently. By testing out the new flow I noticed that this was not necessary anymore. Switching chains works without it

chybisov:
Should we also change the provider here to `account.signer?.provider` or maybe even remove one and use `account.signer?.provider` where needed?

https://github.com/lifinance/widget/blob/f612b21a379034315f97f9adccbb8deea8e450c0/packages/widget/src/providers/WalletProvider/WalletProvider.tsx#L193

Addminus (commented):
i checked it with metamask, walletconnect and taho. seems to work

chybisov (approved):

PR #98, opened 2023-06-07
support ed25519 address

0xksure:
## Background

As of now the `toAddress` is required to be an EVM address. However, as we move towards new non-EVM ecosystems we would need to allow new types of addresses. The first one is to support public keys on the ed25519 elliptic curve. This would enable Solana addresses. 

This change request should serve as a point of discussion and should not be taken as the final solution. 

Additionally, it would make sense to automatically display the `toAddress` field if the `toChain` does not support the EVM address.

### DIFF
--- a/packages/widget/src/hooks/useGetTokenBalancesWithRetry.ts
+++ b/packages/widget/src/hooks/useGetTokenBalancesWithRetry.ts
@@ -1,8 +1,8 @@
-import { isAddress } from '@ethersproject/address';
 import type { Provider } from '@ethersproject/providers';
 import type { Token, TokenAmount } from '@lifi/sdk';
 import { useCallback } from 'react';
 import { useLiFi } from '../providers';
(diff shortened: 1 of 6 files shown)


packages/widget/src/hooks/useGetTokenBalancesWithRetry.ts
```diff
@@ -14,7 +14,7 @@ export const useGetTokenBalancesWithRetry = (provider?: Provider) => {
       depth = 0,
     ): Promise<TokenAmount[] | undefined> => {
       try {
-        const walletAddress = isAddress(accountAddress)
+        const walletAddress = isLiFiAddress(accountAddress)
```
0xksure, line 17: I'm not sure whether `lifi.getTokenBalance` will fail if it receives a non-EVM address. Probably will, so it will have to be a separate task.

chybisov:
Delayed until further notice.

PR #100, opened 2023-06-12
feat: add Safe Wallet to wallet management

Abhikumar98:


packages/wallet-management/src/connectors/safeWalletConnector.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  constructor(args: InjectedConnectorConstructorArgs) {
+    super();
+    this.name = args.name;
+    this.icon = args.icon;
+
+    this.calcAccountData();
+  }
+
+  public installed = () => {
```
Abhikumar98, line 33: added these placeholder methods to keep the interface of the `Wallet` consistent

Addminus: can we add some realy logic here? to detect if its running in the safe app iframe? cause we really do use that function. and if this just returns false, we won't show the wallet

packages/wallet-management/src/wallets.ts
```diff
@@ -228,6 +229,12 @@ const exodus: Wallet = new InjectedConnector(
   (window as any).exodus?.ethereum,
 );
 
+const safe: Wallet = new SafeWalletConnector({
+  name: 'Safe',
+  installed: () => false,
```
Addminus, line 234: same here: can we add some realy logic here? to detect if its running in the safe app iframe? cause we really do use that function. and if this just returns false, we won't show the wallet

Addminus (changes requested):

packages/wallet-management/src/wallets.ts
Abhikumar98: I've made the return type of func `Promise<boolean> | boolean`

packages/wallet-management/src/wallets.ts
```diff
@@ -205,26 +215,28 @@ const tokenpocket: Wallet = new InjectedConnector({
...
+  installed: async () =>
+    (window as any).ethereum?.[ProviderIdentityFlag.Tokenary],
   icon: walletIcons.tokenary,
 });
 
 const exodus: Wallet = new InjectedConnector(
   {
-    name: 'Exodus',
-    installed: () => (window as any).exodus?.ethereum,
+    name: 'ERSLKDJFLKSDJf',
```
chybisov, line 238: 👀

Abhikumar98: hehe, testing local changes 😂

chybisov: I hope you will get it back 😄

packages/wallet-management/src/wallets.ts
```diff
@@ -203,31 +215,44 @@ const tokenpocket: Wallet = new InjectedConnector({
...
   (window as any).exodus?.ethereum,
 );
 
+const safe: Wallet = new SafeWalletConnector({
+  name: 'Safe',
+  installed: async () => {
+    const sdk = new SafeAppsSDK();
+
+    const accountInfo = await sdk.safe.getInfo();
+    return !!accountInfo.safeAddress;
```
chybisov, line 251: Can this throw an error?

packages/wallet-management/src/types/index.ts
```diff
@@ -27,17 +30,17 @@ export interface AccountData {
   chainId: number;
   address: string;
   signer: Signer;
-  provider: Web3Provider;
+  provider: ethers.providers.Web3Provider;
+  isSafeWallet?: boolean;
```
chybisov, line 34: `isMultisigWallet`?

packages/wallet-management/src/types/index.ts
```diff
@@ -27,17 +30,17 @@ export interface AccountData {
...
+  sdk?: SafeAppsSDK;
```
chybisov, line 35: Not sure how to generalize this field, but probably we should try to find a way.

packages/widget/src/App.tsx
```diff
@@ -39,17 +42,25 @@ export const App = forwardRef<WidgetDrawer, WidgetProps>(
...
+        <AppContainer>
+          <Header />
+          <FlexContainer disableGutters>
+            <AppRoutes />
+          </FlexContainer>
+          <PoweredBy />
+          <Initializer />
+        </AppContainer>
+        {expandable ? <RoutesExpanded /> : null}
+      </AppExpandedContainerMultisig>
```
chybisov, line 63: We discussed that this should be on Jumper, please move this component and all related stuff there.

packages/widget/src/types/events.ts
```diff
@@ -6,6 +6,7 @@ export enum WidgetEvent {
   RouteExecutionCompleted = 'routeExecutionCompleted',
   RouteExecutionFailed = 'routeExecutionFailed',
   RouteHighValueLoss = 'routeHighValueLoss',
+  SafeRouteInitiated = 'safeRouteInitiated',
```
chybisov, line 9: If we want a special event when multisig transaction started, we should name it as `MultisigTransactionStarted`. As we discussed, we must avoid using wallet specific terminology here.

chybisov (changes requested):

packages/wallet-management/src/types/index.ts
Abhikumar98: we aren't using the sdk from this place anyways, it's being accessed from the instance present inside the provider. I'll take a look at this and if required remove it

packages/widget/src/i18n/en.json
```diff
@@ -259,5 +259,11 @@
   },
   "wallet": {
     "extensionNotFound": "Please make sure that only the {{name}} browser extension is active before choosing this wallet."
+  },
+  "multisig": {
+    "alert": {
+      "title": "Multisig wallet connected",
+      "description": "Destination wallet address is required to complete the exchange."
+    }
```
chybisov, line 267: Shouldn't be here, move to Jumper, please.

Abhikumar98: Checking

chybisov (approved):

chybisov (approved):

packages/widget/src/components/SelectChainAndToken.tsx
```diff
@@ -36,6 +39,16 @@ export const SelectChainAndToken: React.FC<BoxProps> = (props) => {
...
     !hiddenToToken;
+
+  useEffect(() => {
+    emitter.emit(WidgetEvent.ExecutionPathDetails, {
+      fromChain,
+      fromToken,
+      toChain,
+      toToken,
+    });
+  }, [fromChain, toChain, fromToken, toToken]);
```
chybisov, line 50: We should emit events when we select the token, not inside useEffect. This will emit unnecessary events when mounted.

chybisov (changes requested):

packages/widget/src/components/TokenList/useTokenSelect.ts
```diff
@@ -38,6 +40,20 @@ export const useTokenSelect = (formType: FormType, onClick?: () => void) => {
...
+          chainId: selectedChainId,
+          tokenAddress,
+        });
+      }
+      if (formType === 'to') {
+        emitter.emit(WidgetEvent.DestinationChainTokenSelected, {
+          chainId: selectedChainId,
+          tokenAddress,
+        });
+      }
```
chybisov, line 55: Should we go this way? 🙂

```ts
emitter.emit(
      formType === 'from'
        ? WidgetEvent.SourceChainTokenSelected
        : WidgetEvent.DestinationChainTokenSelected,
      {
        chainId: selectedChainId,
        tokenAddress,
      },
    );
```

packages/widget/src/types/events.ts
```diff
@@ -42,3 +37,8 @@ export interface RouteExecutionUpdate {
   route: Route;
   process: Process;
 }
+
+export interface ExecutionPathDetails {
```
chybisov, line 41: Let's rename it to ChainTokenSelected

chybisov (changes requested):

packages/widget/src/components/SelectChainAndToken.tsx
```diff
@@ -4,9 +4,12 @@ import { useWatch } from 'react-hook-form';
 import { ReverseTokensButton } from '../components/ReverseTokensButton';
 import { SelectTokenButton } from '../components/SelectTokenButton';
 import { FormKey, useWidgetConfig } from '../providers';
-import { DisabledUI, HiddenUI } from '../types';
+import { DisabledUI, HiddenUI, WidgetEvent } from '../types';
+import { useEffect } from 'react';
+import { useWidgetEvents } from '../hooks';
 
 export const SelectChainAndToken: React.FC<BoxProps> = (props) => {
+  const emitter = useWidgetEvents();
```
chybisov, line 12: Please remove it if we no longer use it (imports as well).

chybisov (changes requested):

chybisov (approved):

PR #113, opened 2023-07-12
fix: safe env check logic

Abhikumar98:


packages/wallet-management/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@lifi/wallet-management",
-  "version": "2.2.0",
+  "version": "2.2.1",
```
chybisov, line 3: Why do we increment this?

packages/wallet-management/src/wallets.ts
```diff
@@ -260,12 +260,17 @@ const exodus: Wallet = new InjectedConnector(
 const safe: Wallet = new SafeWalletConnector({
   name: 'Safe',
   installed: async () => {
+    const isIFrameEnvironment = window?.parent !== window;
```
chybisov, line 263: We can return earlier if we are not inside the iframe to prevent 99% of our users from waiting for 200ms 🙂

chybisov (changes requested):

packages/wallet-management/package.json
Abhikumar98: a new version would be deployed when this is merged, right?
incremented for that

packages/wallet-management/src/wallets.ts
Abhikumar98: Done

chybisov (approved):

oktapodia:
Addressed the CHANGES_REQUESTED review in d230c276:

1. **SDK version header** — `@lifi/sdk` does not re-export `version` from its root entry, and the `package.json` import broke the published ESM build. `scripts/version.js` now bakes `sdkVersion` into `src/config/version.ts`; the stream client reads that constant.
2. **Missing `estimate`** — ranking no longer goes through `fees.ts`. Step gas/fees use `step.estimate?.… ?? []`, so a bare `{ tool: 'lifi' }` step no longer throws.
3. **Net USD vs backend** — ported `sumStepGasCostsUsd` / `sumNonIncludedFeeCostsUsd` / `getRouteGasCostUsd` to read `amountUSD` verbatim (including amountUSD-only fixtures with empty `priceUSD` / missing `token`).
4. **`TRUSTED_CUSTOM_PROVIDERS`** — dropped `fynd` and `uniswap` so the set matches backend `SIMULATION_EXCLUDED_CUSTOM_PROVIDERS`. Noted in the docstring that the trusted tier is inert on the progressive SSE path today.
5. **FASTEST docstring** — updated to call out the intentional divergence (skip missing duration; backend still uses `?? 0`).

**Relayer question:** intentional for this POC — with both flags on we keep the batch path so gasless insertion still runs. Merging the relayer quote into the stream result is a follow-up if we want progressive for the mainstream connected case.

PR #116, opened 2023-08-02
ci: switch to pnpm

dnnnvx:
Using Lerna with pnpm wasn't a problem, but since pnpm works a bit differently (doesn't have `node_modules` like `yarn install` does), during `pnpm build` there are inferred types problems, an example:
```
The inferred type of 'FormContainer' cannot be named without a reference to '.pnpm/@mui+system@5.14.3_@emotion+react@11.11.1_@emotion+styled@11.11.0_@types+react@18.2.16_react@18.2.0/node_modules/@mui/system'. This is likely not portable. A type annotation is necessary.
```
(Here's a TypeScript [issue](https://github.com/microsoft/TypeScript/issues/47663) talking about a similar problem)

Those inferred types errors can be manually solved, but not sure if you want that


chybisov:
@dnnnvx can we solve this by adding @mui/system as dependency?

dnnnvx:
> can we solve this by adding @mui/system as dependency?

I'll try

chybisov (commented):
Having so many `any`s is a nightmare 😳

chybisov:
@dnnnvx as I understand, this is not going to fly and we can close it until better times? 🙂

dnnnvx:
> @dnnnvx as I understand, this is not going to fly and we can close it until better times? 🙂

Yes, also other FE repos have some problems migrating to it. I think we should pause it for now and wait and see if it makes sense, since on the BE we still have some open discussion about it, for example: Dependabot doesn't support pnpm, and its developers are not that active

chybisov:
@dnnnvx Thank you for trying this out! 🙂 I'll close it then.

PR #117, opened 2023-08-03
ci: add feature branch actions

dnnnvx:


chybisov (commented):
Can we please change `packages/widget-embedded/dist` to `packages/widget-playground/dist` 🙂

dnnnvx:
> Can we please change `packages/widget-embedded/dist` to `packages/widget-playground/dist` 🙂

Ah I was creating the action for both 👀 should be only playground?

chybisov:
@dnnnvx Yep, `widget-playground` should be enough for now 🙂

chybisov (approved):
Works like a charm 🙂

PR #119, opened 2023-08-14
fix: add close button on mobile

Abhikumar98:
CleanShot 2023-08-14 at 13 46 36@2x

Open to suggestions on the placement of the button 😂


packages/widget/src/AppDrawer.tsx
```diff
@@ -129,3 +135,11 @@ export const DrawerButtonText = ({
...
   );
 };
+
+export const CloseButton = ({ onClick }: { onClick: () => void }) => {
+  return (
+    <CloseButtonLayout onClick={onClick}>
+      <CloseIcon />
+    </CloseButtonLayout>
+  );
+};
```
chybisov, line 145: If this is in the same file anyway, can we please move this to line 111?

packages/widget/src/AppDrawer.style.tsx
```diff
@@ -96,3 +96,24 @@ export const DrawerButtonTypography = styled(Typography)(({ theme }) => ({
   transform: 'rotateZ(180deg)',
   writingMode: 'vertical-rl',
 }));
+
+export const CloseButtonLayout = styled(Box)(
```
chybisov, line 100: There is an IconButton component from MUI, it would be better to use it instead. Also, we don't have bordered icons anywhere in the widget. 🙂

chybisov (dismissed):
Please also check how this works with `split` subvariant

image

Abhikumar98:
CleanShot 2023-09-04 at 19 00 51@2x
CleanShot 2023-09-04 at 19 00 46@2x


Thanks for the suggestions @chybisov, I've made the required change. And additionally keeping the close icon on mobile and desktop as it's the regular drawer behaviour.

packages/widget/src/AppDrawer.style.tsx
Abhikumar98: Done!

Abhikumar98:
> Please also check how this works with `split` subvariant
> 
> image

Yeah, this is causing an issue.
Can you suggest where it'd be best to keep the close button in this variant?

This works well for default and lifuel variant
CleanShot 2023-09-04 at 19 07 10@2x
CleanShot 2023-09-04 at 19 07 05@2x

But breaking for split
CleanShot 2023-09-04 at 19 07 19@2x

Abhikumar98:
I've pushed the icons to the left and accommodated some space for close button, this would happen only if the variant is drawer and subvariant is split

CleanShot 2023-09-04 at 19 31 17@2x
CleanShot 2023-09-04 at 19 31 13@2x

chybisov (changes requested):
@Abhikumar98 I like the idea. There are some small inaccuracies though

1. Icons should be the same size, including hover effects.
image

2. Default mode has unnecessary space on the right of the settings button.
image

Abhikumar98:
> @Abhikumar98 I like the idea. There are some small inaccuracies though
> 
> 1. Icons should be the same size, including hover effects.
> 
> image
> 2. Default mode has unnecessary space on the right of the settings button.
> 
> image

Pushed the changes 😄

chybisov:
@Abhikumar98, please follow the edge 😅
image
image
image

chybisov (changes requested):
@Abhikumar98, please follow the edge 😅

**Before**

image

**After**

image
image

Abhikumar98:
> @Abhikumar98, please follow the edge 😅
> 
> **Before**
> 
> image
> **After**
> 
> image image

Hey @chybisov,
Missed this comment 🙇🏽 

I've pushed the fixes though.
This was really a great notice, thanks for this review!!

PR #121, opened 2023-08-18
fix: remove empty spaces from ens- or walletAddress input (LF-3268)

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-3268


packages/widget/src/components/SendToWallet/SendToWallet.tsx
```diff
@@ -31,6 +31,9 @@ export const SendToWallet: React.FC<BoxProps> = forwardRef((props, ref) => {
     rules: {
       required:
         requiredToAddress && (t('error.title.walletAddressRequired') as string),
+      onChange: (e) => {
+        setValue(FormKey.ToAddress, e.target.value.trim());
```
Abhikumar98, line 35: This looks solid, but I have an very peculiar edge case where the user might paste an ens with a space in middle. To solve for this, we can use a regex to remove all spaces, as `.trim()` also uses a regex afaik internally.

This should work the best, imo
`string.replace(/\s/g, '')`

chybisov: Space in the middle? 👀

Abhikumar98: haha, yeah, not super peculiar edge case, only possible if the user "*wants-wants*" to do it.

chybisov: I mean, are spaces allowed in ENS?

0xsebn: > I mean, are spaces allowed in ENS?

Don't believe they are

chybisov (approved):

Abhikumar98 (approved):

PR #124, opened 2023-08-31
feat: added okx wallet

0xsebn:
Jira: https://lifi.atlassian.net/jira/software/c/projects/LF/boards/6?modal=detail&selectedIssue=LF-4512


packages/wallet-management/src/wallets.ts
```diff
@@ -231,6 +231,15 @@ const tokenary: Wallet = new InjectedConnector({
   icon: walletIcons.tokenary,
 });
 
+const okx: Wallet = new InjectedConnector(
+  {
+    name: 'Okx',
```
chybisov, line 236: I guess the correct name of the brand is `OKX`?

chybisov (changes requested):

PR #130, opened 2023-09-15
fix: custom bridge and exchange select indicator

Abhikumar98:
https://github.com/lifinance/widget/assets/35174288/fdae83b0-b6a7-4433-b39a-c75bec520e92


packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
```diff
@@ -35,25 +44,43 @@ export const ResetSettingsButton: React.FC = () => {
     toggleDialog();
   };
 
+  const isCustomRouteSettings =
+    tools?.bridges?.length !== enabledBridges?.length ||
+    tools?.exchanges?.length !== enabledExchanges?.length;
+
```
chybisov, line 50: We should take into account all settings that can affect route availability, including route priority, gas price, slippage 🙂

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
```diff
@@ -12,8 +12,17 @@ import { Dialog } from '../../components/Dialog';
...
+import { ResetButtonContainer } from './ResetSettingsButton.style';
+import { InfoRounded } from '@mui/icons-material';
 
 export const ResetSettingsButton: React.FC = () => {
+  const { enabledBridges, enabledExchanges } = useSettingsStore((state) => {
+    return {
+      enabledBridges: state.enabledBridges,
+      enabledExchanges: state.enabledExchanges,
+    };
+  });
```
chybisov, line 24: Please use an array and shallow equality function, see other usages of zustand in the app.

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
```diff
@@ -35,25 +44,43 @@ export const ResetSettingsButton: React.FC = () => {
...
+            </DialogContentText>
+          </DialogContent>
+          <DialogActions>
+            <Button onClick={toggleDialog}>{t('button.cancel')}</Button>
+            <Button variant="contained" onClick={handleReset} autoFocus>
+              {t('button.reset')}
+            </Button>
+          </DialogActions>
+        </Dialog>
+      </ResetButtonContainer>
```
chybisov, line 83: We should show message **and** button only if some settings have changed

chybisov (changes requested):

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: Okay, the button should be shown for any settings change whereas the message only for route affecting changes right?

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: okay, checking it out

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: got it, makes sense. I'll include them

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
chybisov: image

All these settings affect routes, so message and button should be shown when we change them. There should be no cases when the button is shown without a message. We can just ignore language and appearance here.

Abhikumar98: Correct, in case of any default widget past configuration modification, I will be reverting back to the configuration passed via the widget

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: I've pushed the changes, I get the importance of `shallow` but a personal opinion, I prefer Object > Array for de-structuring because with objects there isn't any positional values involved, meaning it's differentiated by the unique keys.

But with arrays, wrong positions could lead to incorrect values.

Is there any particular reason to prefer array in a case where we are just reading the values from zustand store?

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: I'm unable to find the type for `gasPrice` in `WidgetConfig` interface, am I missing something here?

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
chybisov: It is the way it works with some other stuff in react like `const [state, setState] = useState()` and also just less code with no need for mapping, etc. Maybe it is also more performant than the object, but I didn't check that.

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
chybisov: Yes, there is no option for `gasPrice` in `WidgetConfig`, but `gasPrice` can be modified by the user, so we also need to consider that.

chybisov:
I don't see any changes to the code 👀

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: Correct, but ideally in react the pattern follow something like the first element is state and the second element is the handler to update the state.

But in this case we are only reading the states from Settings store. Particularly in this scenario, I wanted to know if having it in array helps.

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: Done, pushed the changes

Abhikumar98:
> I don't see any changes to the code 👀

Hey, forgot to push changes, haha.
Pushed now

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
```diff
@@ -31,29 +50,62 @@ export const ResetSettingsButton: React.FC = () => {
...
-          <Button onClick={toggleDialog}>{t('button.cancel')}</Button>
-          <Button variant="contained" onClick={handleReset} autoFocus>
-            {t('button.reset')}
+    isCustomRouteSettings && (
+      <Box px={3} mt={1.5} mb={1}>
+        <ResetButtonContainer>
+          <Box display={'flex'} marginBottom={'12px'}>
+            <InfoRounded
+              sx={{
+                marginRight: '8px',
```
chybisov, line 82: Just a matter of preference and everywhere in the codebase we use `?:` for rendering or in this case it would be nice to do

```ts
if (!isCustomRouteSettings) {
 return null
}

return (<>...</>)
```
You can also search for [Stop using &&](https://kentcdodds.com/blog/use-ternaries-rather-than-and-and-in-jsx) 😄

chybisov (changes requested):

packages/widget/src/stores/settings/useSettingsStore.ts
```diff
@@ -8,9 +8,12 @@ import { SettingsToolTypes } from './types';
 export const defaultConfigurableSettings: Pick<
   SettingsState,
   'routePriority' | 'slippage'
-> = {
+> & {
+  gasPrice: string;
+} = {
   routePriority: 'RECOMMENDED',
   slippage: '0.5',
+  gasPrice: 'slow',
```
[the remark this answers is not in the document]
Abhikumar98: I'll change this to "Normal"

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: hahaha, yeah I read this, the one time I use`&&` 😂.
Thanks for pointing this out though !!

fixing this

packages/widget/src/stores/settings/useSettingsStore.ts
```diff
@@ -8,9 +8,12 @@ import { SettingsToolTypes } from './types';
 export const defaultConfigurableSettings: Pick<
   SettingsState,
   'routePriority' | 'slippage'
-> = {
+> & {
+  gasPrice: string;
+} = {
```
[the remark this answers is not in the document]
Abhikumar98: Missed this, the autocomplete didn't show hence I assumed it's not present. Updated

Abhikumar98:
> Also, I hope you've tested light theme 😄

We need to reset this as well?
I mean, for this we'll hide the message and only show the button right?

Abhikumar98:
> Also, I hope you've tested light theme 😄

https://github.com/lifinance/widget/assets/35174288/5c8aab8a-2e81-46f5-9d92-5bcf56fdb6fa

chybisov:
> > Also, I hope you've tested light theme 😄
> 
> We need to reset this as well? I mean, for this we'll hide the message and only show the button right?

I'm not sure that we need to reset the theme settings - for some users dark theme can be considered as default. 🤔 Let's leave this button only for route stuff for now.

Abhikumar98:
> > > Also, I hope you've tested light theme 😄
> > 
> > 
> > We need to reset this as well? I mean, for this we'll hide the message and only show the button right?
> 
> I'm not sure that we need to reset the theme settings - for some users dark theme can be considered as default. 🤔 Let's leave this button only for route stuff for now.

Alright, removing the reset functionality from theme

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
```diff
@@ -31,29 +50,67 @@ export const ResetSettingsButton: React.FC = () => {
         tools.bridges.map((tool) => tool.key),
         tools.exchanges.map((tool) => tool.key),
       );
+
+      console.log({ config });
```
chybisov, line 54: Please add ESLint and husky rules for this, it happens too often. 😄

chybisov (changes requested):

packages/widget/src/pages/SettingsPage/ResetSettingsButton.tsx
Abhikumar98: About time for this 🥲
Surely I'll do it

chybisov (changes requested):

Abhikumar98:
pushed all the review changes @chybisov

chybisov (approved):

PR #131, opened 2023-09-15
fix: hide browser wallets for mobile and tablets

Abhikumar98:
## When the website is opened in desktop view:

CleanShot 2023-09-15 at 18 26 50@2x

## When the website is opened in mobile/responsive view:
CleanShot 2023-09-15 at 18 26 35@2x

The requirement is to hide the browser wallets and show users the correct option for connecting their wallets via walletconnect.

### DIFF
--- a/packages/widget/src/pages/SelectWalletPage/SelectWalletPage.tsx
+++ b/packages/widget/src/pages/SelectWalletPage/SelectWalletPage.tsx
@@ -20,8 +20,6 @@ import { ListItemText } from '../../components/ListItemText';
 import { useNavigateBack } from '../../hooks';
 import { useWallet } from '../../providers';
 
-const defaultTabWidth = 892;
-
 export const SelectWalletPage = () => {
   const { t } = useTranslation();
   const { navigateBack } = useNavigateBack();
@@ -33,7 +31,7 @@ export const SelectWalletPage = () => {
   const [wallets, setWallets] = useState<Wallet[]>();
 
   const isDesktopView = useMediaQuery((theme: Theme) =>
-    theme.breakpoints.up(defaultTabWidth),
+    theme.breakpoints.up('sm'),
   );
 
   const closeDialog = () => {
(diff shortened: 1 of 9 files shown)


Abhikumar98:
# ⚠️ Do not merge/approve this PR

There is an issue going on with walletconnect on our end (ig), and the resolution of this ticket heavily depends upon Walletconnect.



https://github.com/lifinance/widget/assets/35174288/f2906620-ecdf-4c8c-9473-024ed475c823

packages/widget/src/pages/SelectWalletPage/SelectWalletPage.tsx
```diff
@@ -18,6 +20,8 @@ import { ListItemText } from '../../components/ListItemText';
 import { useNavigateBack } from '../../hooks';
 import { useWallet } from '../../providers';
 
+const defaultTabWidth = 892;
```
chybisov, line 23: Let's reduce it to a more mobile-centric value like the default `sm` breakpoint.

chybisov (changes requested):

packages/widget/src/pages/SelectWalletPage/SelectWalletPage.tsx
Abhikumar98: Sure, we don't want to consider any tablet users?

chybisov: Nice question, tablets resolution can be high to identify it as a tablet or not correctly. They can still select WalletConnect, there are no restrictions on that.

Abhikumar98: Ahh got it. Makes sense

chybisov (approved):

juscamarena:
@chybisov it seems when embedded in a mobile app I don't see this change being triggered, any idea?
wc

PR #139, opened 2023-10-10
fix: update xdefi icon

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-5203

Now with updated icon: 
image

Before:
image


chybisov (approved):

dennyscode:
Oh, well yes. "Fix" would have been better! I will keep that in mind

PR #142, opened 2023-10-12
fix: enable gas sufficiency check for SAFE

Abhikumar98:
For Stargate and Connext, the gas is to be paid from the SAFE account rather than the EOA or it being deducted from the amount being funds being bridged/swapped


packages/widget/src/components/GasMessage/GasMessage.tsx
```diff
@@ -16,7 +16,17 @@ export const GasMessage: React.FC<GasMessageProps> = ({ route, ...props }) => {
   const { sdkConfig } = useWidgetConfig();
   const isMultisigSigner = sdkConfig?.multisigConfig?.isMultisigSigner;
 
-  const validInsufficientGas = insufficientGas?.length && !isMultisigSigner;
+  const selectedTool = route?.steps[0]?.tool;
+  const bridgesForGasWarning = ['connext', 'stargate'];
```
Abhikumar98, line 20: Should I move this a utils/helper file?

chybisov (approved):

PR #148, opened 2023-10-31
feat: add rabby wallet

0xsebn:
jira: https://lifi.atlassian.net/jira/software/c/projects/LF/boards/15?selectedIssue=LF-5463

Will update svg once Vilen posts it on figma

### DIFF
--- a/packages/wallet-management/src/wallets.ts
+++ b/packages/wallet-management/src/wallets.ts
@@ -258,10 +258,10 @@ const okx: Wallet = new InjectedConnector(
 const rabbywallet: Wallet = new InjectedConnector(
   {
     name: 'Rabby Wallet',
-    installed: async () => (window as any).ethereum,
+    installed: async () => (window as any).rabby,
     icon: walletIcons.rabbywallet,
   },
-  (window as any).ethereum,
+  (window as any).rabby,
 );
 const exodus: Wallet = new InjectedConnector(
   {
--- a/packages/wallet-management/src/wallets.ts
+++ b/packages/wallet-management/src/wallets.ts
@@ -255,6 +255,14 @@ const okx: Wallet = new InjectedConnector(
   (window as any).okxwallet,
 );
 
+const rabbywallet: Wallet = new InjectedConnector(
+  {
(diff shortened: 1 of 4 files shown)


0xsebn:
Added some details on the ticket
https://lifi.atlassian.net/jira/software/c/projects/LF/boards/15?assignee=6398600d3c9bcd363977188e&selectedIssue=LF-5463

chybisov:
> Added some details on the ticket https://lifi.atlassian.net/jira/software/c/projects/LF/boards/15?assignee=6398600d3c9bcd363977188e&selectedIssue=LF-5463

We need to find a way to determine when Rabby is installed, or it will compete with MM for window.ethereum. Especially in the new implementation.

packages/wallet-management/src/wallets.ts
```diff
@@ -255,6 +255,14 @@ const okx: Wallet = new InjectedConnector(
   (window as any).okxwallet,
 );
 
+const rabbywallet: Wallet = new InjectedConnector(
+  {
+    name: 'Rabby Wallet',
+    installed: async () => (window as any).ethereum,
```
Addminus, line 261: looked at the wallet injection. for the installed hook
`(window as any).ethereum. isRabby` 
should work

Addminus (changes requested):

packages/wallet-management/src/wallets.ts
Addminus: actually rabby wallet injects also as `window.rabby` we should use that instead of the window.ethereum injection

Addminus: so: installed just checks window.rabby

packages/wallet-management/src/wallets.ts
```diff
@@ -255,6 +255,14 @@ const okx: Wallet = new InjectedConnector(
...
+    icon: walletIcons.rabbywallet,
+  },
+  (window as any).ethereum,
```
Addminus, line 264: when going with window.rabby we should pass `(window as any).rabby)` as injection here

PR #154, opened 2023-11-14
feat: update the settings page

DNR500:
Jira: [LF-5661](https://lifi.atlassian.net/browse/LF-5661)

This presents a reworking of the Settings page - a more detailed description can be found on the ticket. It also introduces a notification badge that is present on the main page if settings have been customised.

Some comments about other considerations for future changes can be found on the ticket too.

SettingsPage-update

[LF-5661]: https://lifi.atlassian.net/browse/LF-5661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/components/Header/SettingsButon.style.tsx
```diff
@@ -0,0 +1,32 @@
+import { darken, styled } from '@mui/material/styles';
+import { IconButton } from '@mui/material';
+import { getInfoBackgroundColor, getWarningBackgroundColor } from '../../utils';
+
+interface SettingsIconButtonProps {
+  notification?: 'info' | 'warning';
+}
+
+export const SettingsIconButton = styled(IconButton, {
+  shouldForwardProp: (props) => props !== 'notification',
```
DNR500, line 10: I figured this out - not having this doesn't cause any console warnings but if the prop you are passing is a string that string will be added to the DOM. For example

without shouldForwardProp

without-shouldForwardProp

and with shouldForwardProp

withShouldForwardProps-button

packages/widget/src/pages/SettingsPage/SlippageSettings/SlippageSettings.style.tsx
```diff
@@ -0,0 +1,93 @@
...
+  borderRadius: theme.shape.borderRadiusSecondary,
+  boxShadow: `0px 2px 4px ${alpha(theme.palette.common.black, 0.04)}`,
+});
+
+interface SlippageDefaultProps {
+  selected?: boolean;
+}
+export const SlippageDefaultButton = styled(ButtonBase)<SlippageDefaultProps>(({
+  theme,
+  selected,
```
DNR500, line 33: I checked the use of the `selected` prop. I think because it is a boolean it doesn't get added to the DOM and so shouldForwardProp isn't needed. Same for the Input below

DNR500:
I've also added the Accordion functionality as a single commit - this one allows one of the expanded panels to be open at once. If we don't like this behaviour we can just drop that commit

chybisov (approved):

Issue #157, opened 2023-11-23
How to update or completely destroy or re-initialize the widget context config ?!


digital-flowers:
We are currently working on integrating a Li-Fi widget into our Quantum wallet. However, we are encountering an issue that requires us to dynamically alter certain widget configurations, specifically the `chains:{allow:[]}` property. Once this property is initially set, there seems to be no way to modify it, even if the widget is completely unmounted.

To provide more context, in our wallet, if a user adds or removes a network, we need to update the chain IDs permitted in the widget. Unfortunately, I haven't found a method to accomplish this.

Furthermore, this issue affects other functionalities in our wallet. Each account is associated with different networks, and when we switch accounts, the chain IDs are altered but not reflected in your widget.

Here is my code:

```javacript
const chainIds = useAccountChainIds();

const config:WidgetConfig = {
      integrator: "quantum",
      variant: "default",
      subvariant: "default",
      subvariantOptions: "swap",
      appearance: "light",
      hiddenUI: ["language", "appearance", "history"],
      languages: {
        default: "en",
      },
      chains: {
        allow: chainIds,
      },
      walletManagement: signer ? {
            signer,
            connect: async () => {
              return signer;
            },
            disconnect: async () => {},
          }
        : undefined,
    }
};

return <LiFiWidget integrator={widgetConfig.integrator} config={widgetConfig} />;

```

and here are the use case:

**Account 2:**

<img width="150" alt="Screenshot 2023-11-24 at 2 00 35 AM" src="https://github.com/lifinance/widget/assets/5216291/da28ce1b-f716-4ba8-bb74-99c7b324d019">

<img width="150" alt="Screenshot 2023-11-24 at 2 00 48 AM" src="https://github.com/lifinance/widget/assets/5216291/eb543b69-e1c2-48ee-9d7a-7e85df8ed24d">

Switch to
 **Account 1:**

<img width="150" alt="Screenshot 2023-11-24 at 2 01 01 AM" src="https://github.com/lifinance/widget/assets/5216291/9ed7b734-5be7-4555-97d9-d04b70220dcb">

<img width="150" alt="Screenshot 2023-11-24 at 2 01 11 AM" src="https://github.com/lifinance/widget/assets/5216291/9b2da81f-f1ca-44d4-be46-f2ba9ebf9ff1">

chybisov:
@digital-flowers hey, it should be fixed in `v2.9.4`. Please try it 🙂

PR #158, opened 2023-11-27
fix: enable chains update from config

DNR500:
Jira: [LF-5915](https://lifi.atlassian.net/browse/LF-5915)

This attempts to address this issue

https://github.com/lifinance/widget/issues/157

Adds chain to the queryKey so a refetch will be triggered if chain in the config changes

See the ticket if you want assistance to set this up locally to test

[LF-5915]: https://lifi.atlassian.net/browse/LF-5915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/hooks/useChains.ts
```diff
@@ -17,34 +17,36 @@ export const useChains = () => {
...
-      if (!toChainValue) {
-        setValue(FormKey.ToChain, chainOrder[0]);
-      }
-      return filteredChains;
-    },
-    {
-      enabled: Boolean(availableChains),
-    },
-  );
+  const filteredChains = useMemo(() => {
```
DNR500, line 20: I've switch this to useMemo which appears to work. Would be good to have a conversation just to check if there was a reason that useQuery was being used previous for a specific reason or not.

packages/widget/src/hooks/useChains.ts
```diff
@@ -17,34 +17,36 @@ export const useChains = () => {
...
+    ]);
+    if (!fromChainValue) {
+      setValue(FormKey.FromChain, chainOrder[0]);
+    }
+    if (!toChainValue) {
+      setValue(FormKey.ToChain, chainOrder[0]);
+    }
+    return filteredAvailableChains;
+  }, [
+    keyPrefix,
```
DNR500, line 43: Could do with double checking and understanding the use of keyPrefix here too, I've added it to the dep array as it was used in the queryKey previously

packages/widget/src/hooks/useChains.ts
DNR500: We are using the useQuery hook to provide caching across the app so we don't have to keep regenerating the list. Will move this back tot he previous commit

packages/widget/src/hooks/useChains.ts
DNR500: keyPrefix is used in the config for when you might have multiple instance of the widget running in the same app

DNR500:
I'm going to drop the useMeo approach and look at updating the useQuery querykey approach instead

packages/widget/src/hooks/useChains.ts
```diff
@@ -11,21 +11,32 @@ export const useChains = () => {
...
       );
+
+      return filteredChains;
+    },
+    {
+      enabled: Boolean(availableChains),
+    },
+  );
+
+  useEffect(() => {
```
DNR500, line 38: So had a problem in the UI where it there is an attempt to reset the chains to the default settings of an empty array or undefined that the following UI can happen where there are 4 missing chains (there should be the full 18 in teh image below).

issue-rendering chains

Only way I've found so far to get around this is to use a useEffect block to make sure the zustand. Going to see if there is a function on React Queries useQuery hook that might work as well

DNR500: I'm not finding any ting in relation to useQuery thats going to be helpful and I the useEffect will hav to stay if we want to avoid the problem pictured here

DNR500:
On commit https://github.com/lifinance/widget/pull/158/commits/9e845c43e7ccc26e37d8bd69fd6f3b7dc1a506ee I've spiked an approach that does reduce number calls to the useEffect block that sets the chainOrder - previously it could be fired 12 or 15 times on the page but now it seems it fires a max of two times. I think this can still be improved and cleaned up a little. I'll attempt to look at that on the the next commit.

chybisov (approved):

PR #166, opened 2023-12-11
feat: removal of react hook form

DNR500:
Jira: [LF-5981](https://lifi.atlassian.net/browse/LF-5981)

This PR looks to remove the react-hook-form library from the Widget

It puts a zustand store in its place to handle the passing of form values across the app.

Main hooks that interact with the form store

- **useFieldValue** - now provides access to the form values
- **useFieldActions** - provides functions to update the form
- **useFieldController** - provides an interface with onChange, onBlur and value to hook up to inputs

other hooks
- **useTouchedFields** - provides a map which fields are touched
- ***useValidation** - provides access to isValid and errors
- ***useValidationActions** - functions to set up and trigger the validatoiin

*The validation here is a placeholder and will likely change when looking at the new changes for the Wallet Switcher (should be the next piece of work)

[LF-5981]: https://lifi.atlassian.net/browse/LF-5981?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


DNR500:
We might want to revise what hooks are made available for the external API?

see commit https://github.com/lifinance/widget/pull/166/commits/ecba2d5b9fbc2b5b83cb6e81ee9288c2d434cec2

packages/widget/src/types/index.ts
```diff
@@ -1,3 +1,10 @@
 export * from './events';
 export * from './token';
 export * from './widget';
+
+// TODO: needs to be removed
+declare global {
+  interface Window {
+    useFormStore?: boolean;
+  }
+}
```
chybisov, line 10: Can we remove this?

packages/widget/src/stores/form/createFormStore.ts
```diff
@@ -0,0 +1,227 @@
...
+        set(() => ({
+          userValues,
+          touchedFields,
+        }));
+      },
+      getFieldValues: (...names) =>
+        names.map((name) => get().userValues[name]?.value),
+
+      // The following a is a "placeholder" "for validation for toAddress
+      // is will be re-addressed in the next piece of work for Wallet Switching
```
chybisov, line 173: Can we please improve this comment? 😄

packages/widget/src/stores/form/createFormStore.ts
```diff
@@ -0,0 +1,227 @@
...
+
+      return {
+        ...accum,
+        [key]: formValue,
+      };
+    },
+    { ...valuesToFormValues(formDefaultValues) },
+  );
+
+// TODO: isValidating & isValid
```
chybisov, line 75: I guess this is done?

packages/widget/src/components/SendToWallet/SendToWallet.tsx
```diff
@@ -38,44 +50,52 @@ export const SendToWallet: React.FC<BoxProps> = forwardRef((props, ref) => {
...
+        const address = await getEnsAddress(config, {
+          chainId: getFieldValues('toChain')[0],
+          name: normalize(value),
+        });
+        return Boolean(address);
+      } catch (error) {
+        return t('error.title.walletEnsAddressInvalid') as string;
+      }
+    });
+  }, [addFieldValidation, requiredToAddress, getFieldValues, t, config]);
```
chybisov, line 72: Should we add this field onBlur instead of creating validation on every field update?

packages/widget/src/hooks/useDebouncedWatch.ts
```diff
@@ -1,10 +1,8 @@
 import { useEffect, useRef, useState } from 'react';
-import { useWatch } from 'react-hook-form';
+import { useFieldValues } from '../stores';
 
 export const useDebouncedWatch = (name: any, delay: number) => {
```
chybisov, line 4: Can we remove `any` and type `name` properly here now?

packages/widget/src/stores/form/FormUpdater.tsx
```diff
@@ -0,0 +1,69 @@
...
+    chains,
+    fromChain,
+    toChain,
+    isTouched,
+    resetField,
+    setFieldValue,
+  ]);
+
+  // Makes widget config options reactive to changes
+  // Acts similar to values property from useForm, but includes additional logic for chains
```
chybisov, line 50: Should be updated I guess, we no longer have `useForm`

packages/widget/src/stores/form/types.ts
```diff
@@ -0,0 +1,134 @@
+import type { UseBoundStoreWithEqualityFn } from 'zustand/esm/traditional';
+import type { StoreApi } from 'zustand/esm';
```
chybisov, line 2: I think ESM is used by default, so we can safely import from `zustand` and `zustand/traditional`.

chybisov (changes requested):
I couldn't find URLSearchParamsBuilder event though we export it from form barrel file.

chybisov:
> We might want to revise what hooks are made available for the external API?
> 
> see commit [44b71ec](https://github.com/lifinance/widget/commit/44b71ec8b9c6e1dea6520886eeb9fbc134237019)

Maybe later, those are undocumented exports for our experimental NFT flow.

packages/widget/src/types/index.ts
DNR500: Oh yes. I'll get rid of that

packages/widget/src/stores/form/types.ts
DNR500: I'll check those and change them

packages/widget/src/stores/form/createFormStore.ts
DNR500: Actually we can probably live without it - I'll delete it

packages/widget/src/hooks/useDebouncedWatch.ts
DNR500: Actually yes and I think the use of the useDebouncedWatch probably doesn't need to be used with FormKey anymore

packages/widget/src/stores/form/FormUpdater.tsx
DNR500: Will update

packages/widget/src/hooks/useDebouncedWatch.ts
```diff
@@ -1,19 +1,17 @@
 import { useEffect, useRef, useState } from 'react';
-import { useWatch } from 'react-hook-form';
+import { FormFieldNames, useFieldValues } from '../stores';
 
-export const useDebouncedWatch = (name: any, delay: number) => {
-  const watchedValue = useWatch({
-    name,
-  });
+export const useDebouncedWatch = (name: FormFieldNames[], delay: number) => {
+  const watchedValue = useFieldValues(...name);
```
chybisov, line 5: Can we pass `name` instead of `...name` here?

packages/widget/src/components/SendToWallet/SendToWallet.tsx
DNR500: We discussed this and decide for now its actually ok for it to stay here as it only really will update on requiredToAddress and config changing - though we might look to optimise the pattern for validation in future.

When testing this is wasn't actually working. This was main due to a problem in the useValidationActions hook and a missing call to clearErrors when valid. 

Now fixed and works as it did previously  - the additional clearErrors logic has been put in the formStore logic and the useValidationActions has been rewritten like the useFieldActions hook so that it avoid any potential problems related to connascence of position/meaning. See commit https://github.com/lifinance/widget/pull/166/commits/9ebef7fd5b78e219d1aeb961b070ac654fc47c14

DNR500:
> I couldn't find URLSearchParamsBuilder event though we export it from form barrel file.

I've fixed the gitignore file now, this was filtering out the URLSearchParamsBuilder.tsx file and listing it as ignored due to **build**. The change to the gitignore file also changes how dist folders are ignored. I've checked that this works locally.

I think this should be ready for review again

chybisov (approved):

PR #171, opened 2024-01-18
feat: add reduce amount button

chybisov:
image


packages/widget/src/components/Card/CardButton.tsx
```diff
@@ -0,0 +1,18 @@
+import { Button } from '@mui/material';
+import { alpha, styled } from '@mui/material/styles';
+
+export const CardButton = styled(Button)(({ theme }) => {
```
DNR500, line 4: This will cause a conflict I think with the CardButton that is currently on the Bookmark PR. I can change the name on other PR if you like will connect with you tomorrow :-)

PR #172, opened 2024-01-25
feat: curated wallets for the destination address selection

DNR500:
Jira: [LF-5324](https://lifi.atlassian.net/browse/LF-5324)

Introducing the Curated Wallets section that as a follow on to the bookmarked send to wallets work

### Testing

This heavy relies on changes to the widget config. You should see the entries below commented out in the widget-playground

```
toAddress: {
    name: 'Jenny',
    address: '0xAB3Afc314e75dC1648A2E31c06861A07a048C050',
    chainType: ChainType.EVM,
  },
toAddresses: [
  {
    name: 'Lenny',
    address: '0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea9',
    chainType: ChainType.EVM,
  },
  {
    address: '0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea0',
    chainType: ChainType.EVM,
  },
  {
    name: 'Sol',
    address: '6AUWsSCRFSCbrHKH9s84wfzJXtD6mNzAHs11x6pGEcmJ',
    chainType: ChainType.SVM,
  },
],
```
`toAddress` and `toAddresses` work independently of each other 

- adding `toAddresses` will be displayed as the default on the Main page
- adding `toAddresses` will enable the SendToConfiguredWallet page in place of SendToWallet page and display a selectable list of addresses that are featured in the `toAddresses` array - click these addresses should populate the toAddress field on the main page

As part of the this work `toAddress` as feature in the query string parameters will also be supported, for example

`http://localhost:3000/?toAddress=0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea9`

This address should be displayed on the main page in the Send to Wallet section.

We do some checks to try and present a name for that address if we have it.

- Firstly we check to see if the address matches an address listed in the configs `toAddress`
- if that doesn't match we try to find a matching address in the `toAddresses` from the config
- if that doesn't match then we try to find a matching address from the users Bookmarks in localstorage

If non of those match we just display the address

[LF-5324]: https://lifi.atlassian.net/browse/LF-5324?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/widget/src/providers/WidgetProvider/WidgetProvider.tsx
+++ b/packages/widget/src/providers/WidgetProvider/WidgetProvider.tsx
@@ -4,6 +4,7 @@ import { version } from '../../config/version';
 import { setDefaultSettings } from '../../stores';
 import { formatInputAmount, getChainTypeFromAddress } from '../../utils';
 import type { WidgetContextProps, WidgetProviderProps } from './types';
+import type { WidgetConfig } from '../../types';
 
 const initialContext: WidgetContextProps = {
   elementId: '',
@@ -16,6 +17,30 @@ export const useWidgetConfig = (): WidgetContextProps =>
   useContext(WidgetContext);
 
 let sdkInitialized = false;
+
+const attemptToFindMatchingToAddressInConfig = (
+  address: string,
+  config: WidgetConfig,
+) => {
+  if (config.toAddress && config.toAddress.address === address) {
+    return config.toAddress;
+  }
+
+  if (config.toAddresses?.length) {
+    const matchingToAddress = config.toAddresses.find(
+      (toAddress) => toAddress.address === address,
+    );
+
+    if (matchingToAddress) {
+      return matchingToAddress;
+    }
+  }
+
+  return {
+    address: address,
+    chainType: getChainTypeFromAddress(address),
+  };
+};
 
 export const WidgetProvider: React.FC<
   React.PropsWithChildren<WidgetProviderProps>
@@ -60,13 +85,11 @@ export const WidgetProvider: React.FC<
           !isNaN(parseFloat(searchParams.fromAmount))
             ? formatInputAmount(searchParams.fromAmount)
             : widgetConfig.fromAmount,
-        //  TODO: Question: is using getChainTypeFromAddress ok here? Or do we have to start output
-        //   the name and the chainType to the querystring?
         toAddress: searchParams.toAddress
-          ? {
-              address: searchParams.toAddress,
-              chainType: getChainTypeFromAddress(searchParams.toAddress),
-            }
+          ? attemptToFindMatchingToAddressInConfig(
+              searchParams.toAddress,
+              widgetConfig,
+            )
           : widgetConfig.toAddress,
         integrator: widgetConfig.integrator,
         elementId,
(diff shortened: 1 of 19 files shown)


packages/widget/src/providers/WidgetProvider/WidgetProvider.tsx
```diff
@@ -17,6 +18,30 @@ export const useWidgetConfig = (): WidgetContextProps =>
 
 let sdkInitialized = false;
 
+const attemptToFindMatchingToAddressInConfig = (
```
DNR500, line 21: Here we do the query param toAddress check agains the `toAddress` and `toAddresses` in config values.  We don't have access to the Bookmarks from local storage at this point so we have to do checks against  Boolmark later in the code execute - see the onRehydrateStorage function in createBookmarkStore

chybisov: Looks good. Can we please create `utils.ts` in this folder and move this helper function there? 🙂

packages/widget/src/stores/bookmarks/createBookmarkStore.ts
```diff
@@ -57,6 +63,16 @@ export const createBookmarksStore = ({ namePrefix }: PersistStoreProps) =>
...
         }),
+        onRehydrateStorage: () => {
+          return (state) => {
+            if (state && toAddress && !toAddress.name) {
+              const existingBookmark = state.getBookmark(toAddress.address);
+              if (existingBookmark) {
+                state.setSelectedBookmark(existingBookmark);
+              }
+            }
+          };
```
chybisov, line 74: Yesterday we tried and seemed to work with the main function, not return one 👀 Did you find something?

DNR500: I double checked it. The main function doesn't have the bookmarks from local storage  - the bookmarks are only populated in the returned function. Might have been outputting the wrong things in the console yesterday?? When moving the functionality to the returned function it works 👍

packages/widget/src/providers/WidgetProvider/WidgetProvider.tsx
```diff
@@ -4,6 +4,7 @@ import { version } from '../../config/version';
 import { setDefaultSettings } from '../../stores';
 import { formatInputAmount } from '../../utils';
 import type { WidgetContextProps, WidgetProviderProps } from './types';
+import { attemptToFindMatchingToAddressInConfig } from '../../providers/WidgetProvider/utils';
```
chybisov, line 7: I think it should be imported `./utils` 🙂

packages/widget/src/stores/routes/utils.ts
```diff
@@ -6,20 +6,14 @@ export const isRouteDone = (route: RouteExtended) => {
 };
 
 export const isRoutePartiallyDone = (route: RouteExtended) => {
-  return route.steps.some(
-    (step) =>
-      step.execution?.process.some(
-        (process) => process.substatus === 'PARTIAL',
-      ),
+  return route.steps.some((step) =>
+    step.execution?.process.some((process) => process.substatus === 'PARTIAL'),
```
chybisov, line 10: Do you have a different code formatting tool? We should use the same prettier setup and not change existing code formatting 🙂

DNR500: I'll check but we should be using the same prettierrc file

chybisov (approved):

PR #176, opened 2024-01-29
feat: widget next playground

DNR500:
Work in Progress


packages/widget-next-playground/.eslintrc.json
```diff
@@ -0,0 +1,4 @@
+{
+  "root": true,
```
DNR500, line 2: I'll continue to look aligning the eslint to the other projects once we have this merged in

chybisov:
Let's make these two colors match so we use the same primary color for the playground as we set for the widget and the same for the secondary if any 🙂 
image

packages/widget-next-playground/src/app/components/DrawerControls/DesignControls/DesignControls.style.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+export const ColorInput = styled(InputBase)<InputBaseProps>(
+  ({ theme, value }) => ({
+    position: 'relative',
+    border: 'none',
+    height: '100%',
+    width: 97,
+    padding: 0,
+    backgroundColor: value as string,
+    borderRadius: theme.shape.borderRadiusSecondary,
```
chybisov, line 141: Should we apply the same logic as we have for Tabs to make the radius match the container radius proportionally?

```suggestion
    borderRadius: Math.max(theme.shape.borderRadius, theme.shape.borderRadiusSecondary) - 4,
```

### Before

image

### After

image

chybisov:
We should store the drawer state (opened/closed) in localStorage (hi, zustand) to remember it between page refreshes. 😄

packages/widget-next-playground/.env.local
```diff
@@ -0,0 +1 @@
+NEXT_PUBLIC_VITE_WALLET_CONNECT=5432e3507d41270bee46b7b85bbc2ef8
```
chybisov, line 1: Could we simplify this by removing the VITE part, as this is no longer a vite project? 🙂

packages/widget-next-playground/src/app/defaultWidgetConfig.ts
```diff
@@ -0,0 +1,20 @@
+import type { WidgetConfig } from '@lifi/widget';
+
+// TODO: question do we want to move some of the sections of code we use
+//  for testing in the other playgrounds config into here?
```
chybisov, line 4: Yes, please 😄

chybisov (changes requested):

packages/widget-next-playground/.env.local
DNR500: 😂 I can't believe I didn't see that. Will remove it 👍

DNR500:
I did look into a warning that we get in the console when running the widget in Next.js

`bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)`

It will still work with this warning. One of the best posts I could find for this was...

https://solana.stackexchange.com/questions/4077/bigint-failed-to-load-bindings-pure-js-will-be-used-try-npm-run-rebuild-whe

According to top answer there this is bigint falling back to using a js implementation rather than using the C bindings - its means it will still work though it might not be as performant as the when using the C implementation.

To get the C version working you need to have python, make and GCC available locally when running a build (so it is partly environmental). Some of the anwers there also suggest installing bigint-buffer as part of the project to ensure the C bindings are available locally.

I tried all these things but could not consistently get the production build of the next project to run without this warning. It may not be as performant as with the native C bindings but it will not break anything.

chybisov:
@DNR500, thank you for doing the research! I think we can live with it 🙂

chybisov (approved):
~~LGTM!~~ LFG! 🚀😄

PR #179, opened 2024-01-31
feat: display paid bridge fees in transaction history

chybisov:


DNR500 (approved):
The code looks good 👍

PR #181, opened 2024-02-05
feat: add popular tokens config option

chybisov:
Test with BSC chain


DNR500 (approved):
Tested locally. Nice one! 🙂

PR #183, opened 2024-02-07
refactor: improve ESM support and add remix example

chybisov:


DNR500 (approved):
Code looks clean. I think a few barrel files could do with being removed but I'm happy for this to be merged

PR #200, opened 2024-02-26
style: additional theme refinements

chybisov:
- Better handling of avatar badges cut out now using CSS Mask
- Unified list item button hover colors
- Adjusted the size of the clear button for the destination wallet
- Added a new tertiary button variant (used for Show All routes and as Done button for destination wallet)


DNR500 (approved):
Looks good

PR #201, opened 2024-02-26
feat: playground code section

DNR500:
Jira: [LF-2410](https://lifi.atlassian.net/browse/LF-2410)

This adds the Code section to the Playground

[LF-2410]: https://lifi.atlassian.net/browse/LF-2410?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget-playground/src/components/DrawerControls/CodeControl/CodeEditor.tsx
```diff
@@ -0,0 +1,176 @@
...
+    ? configTemplate(
+        configToStringWithSubstitions(getWhitelistedConfig(config)),
+      )
+    : null;
+
+  useEffect(() => {
+    if (code) {
+      setEditorContent(code);
+    }
+  }, [code]);
```
chybisov, line 70: This `useEffect` part looks unnecessary and maybe we can use `code` just fine and avoid re-render 🙂 Wdyt?

packages/widget-playground/src/components/DrawerControls/CodeControl/CodeEditor.tsx
```diff
@@ -0,0 +1,176 @@
...
+        PopperProps={{ style: { zIndex: tooltipPopperZIndex } }}
+        arrow
+      >
+        <CodeCopyButton onClick={handleCopyCode}>
+          <ContentCopyIcon fontSize={'small'} />
+        </CodeCopyButton>
+      </Tooltip>
+      <EditorContainer ref={editorContainerRef}>
+        <Editor
+          loading={<CircularProgress />}
```
chybisov, line 153: Would be nice to use <Skeleton /> here instead of spinner 🙂

chybisov (changes requested):

packages/widget-playground/src/components/DrawerControls/CodeControl/CodeEditor.tsx
DNR500: Have remove this and passed the code directly through to the CodeEditor

packages/widget-playground/src/components/DrawerControls/CodeControl/CodeEditor.tsx
DNR500: Added a Skeleton component instead

chybisov (approved):

PR #208, opened 2024-03-11
feat: improve widget customization

chybisov:


packages/widget/src/themes/createTheme.ts
```diff
@@ -0,0 +1,415 @@
+import { loadingButtonClasses } from '@mui/lab';
```
DNR500, line 1: I really like having the theme unified and present like this in the one function :-)

packages/widget/src/themes/jumper.ts
```diff
@@ -0,0 +1,37 @@
+import type { WidgetTheme } from '../types/widget.js';
+
+export const jumperTheme: WidgetTheme = {
```
DNR500, line 3: This theme doesn't seem to do a lot - is it still needed?

packages/widget/src/themes/windows95.ts
```diff
@@ -0,0 +1,98 @@
...
+        root: {
+          backgroundColor: '#fff',
+          boxShadow:
+            'inset -1px -1px #ffffff, inset 1px 1px #0a0a0a, inset -2px -2px #dfdfdf, inset 2px 2px #808080',
+        },
+      },
+    },
+    MuiIconButton: {
+      defaultProps: {
+        // disableRipple: true,
```
DNR500, line 82: Should we need to keep this?

DNR500 (approved):
This looks good. Couple of questions about if things need deleting but nothing that would stop this being merged in. Thanks Eugene :-)

packages/widget/src/themes/jumper.ts
chybisov: I'll update it later this week 🙂

PR #211, opened 2024-03-12
feat: add theme controls to playground

DNR500:
Jira: [LF-7338](https://lifi.atlassian.net/browse/LF-7338)

This introduces support for base themes into the palyground. These can be a starting point for users to work from to create there own custom themes

Screenshot 2024-03-13 at 19 14 41

[LF-7338]: https://lifi.atlassian.net/browse/LF-7338?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget-playground/src/store/widgetConfig/utils/getConfigOutput.ts
```diff
@@ -0,0 +1,19 @@
+import type { WidgetConfig, WidgetTheme } from '@lifi/widget';
+
+export const getConfigOutput = (
+  config: Partial<WidgetConfig>,
+): Partial<WidgetConfig> => {
+  const { playground, ...theme } = config.theme as WidgetTheme;
```
DNR500, line 6: filtering out the playground property here before output it to the code editor - but I just want to double check that this is purely used for the playground and will never be used in the widget itself?

packages/widget-playground/src/store/editTools/themes.ts
```diff
@@ -0,0 +1,13 @@
+import { windows95Theme } from '@lifi/widget';
+import type { ThemeItem } from './types';
+
+export const themeItems: ThemeItem[] = [
+  {
+    id: 'windows95',
+    name: 'Windows 95',
+    theme: windows95Theme,
+    options: {
+      restrictAppearance: 'light',
```
DNR500, line 10: this setting restricts the mode/appearance in the playground. I've also added an indicator that this is restricted and message in the the Appearance control.

packages/widget-playground/src/components/DrawerControls/DesignControls/ColorControls.tsx
```diff
@@ -1,18 +1,45 @@
...
+  'grey 400': 'theme.palette.grey.400',
+  'grey 500': 'theme.palette.grey.500',
+  'grey 600': 'theme.palette.grey.600',
+  'grey 700': 'theme.palette.grey.700',
+  'grey 800': 'theme.palette.grey.800',
+  'grey 900': 'theme.palette.grey.900',
+  'grey A100': 'theme.palette.grey.A100',
+  'grey A200': 'theme.palette.grey.A200',
+  'grey A400': 'theme.palette.grey.A400',
+  'grey A700': 'theme.palette.grey.A700',
```
DNR500, line 42: These will only appear in the color control in the playground if there are featured in the theme

packages/widget-playground/src/components/DrawerControls/DesignControls/ColorControls.tsx
```diff
@@ -44,11 +71,11 @@ const ColorSelector = ({
 
   return colorValue ? (
     <ColorSelectorContainer {...rest}>
-      {colorName}
+      <CapitalizeFirstLetter>{colorName}</CapitalizeFirstLetter>
       <ColorInput
         aria-label={`${colorName} color selection`}
         type="color"
-        value={colorValue}
+        value={safe6DigitHexColor(colorValue)}
```
DNR500, line 78: The browsers color input doesn't like 3 digit hex colors. This does a safe conversion if they are featured

packages/widget-playground/src/store/widgetConfig/utils/getConfigOutput.ts
chybisov: Yes, that property is for playground only. For now, it has only a background color for the widget, but maybe we will extend it with something else.

packages/widget-playground/src/components/DrawerControls/DesignControls/ColorControls.tsx
chybisov: Maybe it is better to agree on using 6-digit values in the themes?

DNR500: Maybe moving foward thats a good idea. If we ever get to making the config updatable we will likely have to think about doing things like this as well

chybisov (approved):

PR #216, opened 2024-03-22
docs: update examples of widget

DNR500:
Jira: [LF-7415](https://lifi.atlassian.net/browse/LF-7415)

Ensuring that all the examples are on version 3 of the widget

- Some polyfills added
- Other deps updated
- Older Next.js example removed (as it provided nothing that wasn't usable from the Next14 project)
- Nuxt example added
- Also covers https://lifi.atlassian.net/browse/LF-5420

[LF-7415]: https://lifi.atlassian.net/browse/LF-7415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


examples/gatsby/gatsby-node.js
```diff
@@ -0,0 +1,16 @@
+const webpack = require('webpack');
```
DNR500, line 1: I tried doing this with a ts file but couldn't get it to work so have fallen back to using a js file here which does seem to work

PR #234, opened 2024-04-11
chore: add import sort order to eslint

DNR500:
Jira: [LF-7839](https://lifi.atlassian.net/browse/LF-7839)

Use the [eslint-plugin-simple-import-sort](https://www.npmjs.com/package/eslint-plugin-simple-import-sort) package.

This should run on huskys pre-commit hook and will autofix the imports so that we don't have to think about it too much it.

I've run the linter on this codebase as part of the PR so you can see what the output is on our current imports.

Main changes to out current imports are 
- alphabetically sorting 
- Anything that doesn't output a variable as part of the import statement will be moved to the top


DNR500:
Closing this PR as have switched to looking at prettier as an option - see https://github.com/lifinance/widget/pull/238

PR #235, opened 2024-04-15
spike: get eslint v9 working

DNR500:
I’ve been hacking away at tying to get the eslint v9 set up working In the widget codebase - specifically I was spiking out the new Eslint in the widget package listing again.

Main problem I was having was with `eslint-config-react-app` and I can’t get that to work at all  - not even when using with using some of the tooling eslint provides to help migrate legacy config.

Some of the other config and plugin tools have been updated - for example `eslint-plugin-prettier` has documentation for the new “flat config” style for eslint and its easy to get those working.

I couldn’t find anything about ‘‘eslint-config-react-app’ using the new eslint flat config - its not been published in a couple of years and we might have to seriously think about moving away from it. I started to take a look at the plugins and config it uses under the hood - some have already adopted support for flat config, such as `eslint-plugin-react`, but others are still implementing support, for example `eslint-plugin-react-hooks` is maintained by the React team and they still have [an open PR for supporting the new config style for eslint](https://github.com/facebook/react/pull/28773).


DNR500:
This was just created to record my findings and does need to be merged in so I'm closing it

PR #238, opened 2024-04-17
chore: add prettier plugin to sort imports

DNR500:
Jira: [LF-7839](https://lifi.atlassian.net/browse/LF-7839)

Applies sorting using prettier - which means that if anyone runs prettier on save will get instant sorting of imports

I have ended up adding the format command to the pre-commit hook script execution so that even if anyone commits without first running prettier imports should be sorted to the correct order - this will impact the speed of the pre-commit validation.

I explored two plugin options on this PR the first one was trying to use `@trivago/prettier-plugin-sort-imports`. The second was `@ianvs/prettier-plugin-sort-imports`

`@trivago/prettier-plugin-sort-imports` matched the alphabetic that matches what VSCode currently does but it also moves around "side effect" - this isn't a good thing and in some circumstances can actually break things.

So I moved to using `@ianvs/prettier-plugin-sort-imports` instead. This doesn't move the side effect imports and is therefore a little bit safer. Its sorting of named variables tends to case insensitive thought and not match the VSCode way of doing things - I wonder if as we are using prettier and this would be consistent across all IDE's anyway that this would be less important? 

Last prettier plugin I tried was `prettier-plugin-organize-imports` - this one doesn't take into account side effects but I think its the closest to the vscode formatting. Seem to do the minimal changes with some packages remaining unchanged by running the format command.

I've retained the code changes for each plugin as commits on this branch so that changes that each prettier plugin can be looked at and compared in terms of output. Probably good to have a discussion to decide if we are happy to use `prettier-plugin-organize-imports`.

[LF-7839]: https://lifi.atlassian.net/browse/LF-7839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


chybisov (approved):
LFG 🚀

DNR500:
Updated the yarn lock file this should be good to merge

PR #249, opened 2024-05-21
fix: display of name for toAddress in the playground

DNR500:
jira: [LF-8293](https://lifi.atlassian.net/browse/LF-8293)

This PR adds a diff and patch to the rehydrate function to the config to make sure that the toAddress data will be passed correctly to the widget


packages/widget-playground-vite/package.json
```diff
@@ -44,7 +44,7 @@
     "source-map-explorer": "^2.5.3",
     "typescript": "^5.4.5",
     "vite": "^5.2.11",
-    "vite-plugin-node-polyfills": "^0.21.0",
+    "vite-plugin-node-polyfills": "^0.22.0",
```
DNR500, line 47: Had to bump this as it seemed to be causing problems with the Vite chunk optimisation

chybisov (approved):

PR #251, opened 2024-05-22
feat: add onRoutesAvailable widget-event

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-8355

### DIFF
--- a/packages/widget/src/components/Routes/RoutesExpanded.tsx
+++ b/packages/widget/src/components/Routes/RoutesExpanded.tsx
@@ -105,9 +105,16 @@ export const RoutesExpandedElement = () => {
   );
 
   const routeNotFound = !currentRoute && !isLoading && !isFetching && expanded;
-
+  const routesAvailable = (routes && routes?.length > 0) || false;
   const toAddressUnsatisfied = currentRoute && requiredToAddress && !toAddress;
   const allowInteraction = account.isConnected && !toAddressUnsatisfied;
+
+  useEffect(() => {
+    !isLoading &&
+      !isFetching &&
+      isFetched &&
+      emitter.emit(WidgetEvent.RoutesAvailable, routesAvailable);
+  }, [routesAvailable, emitter, isLoading, isFetching, isFetched, routes]);
 
   useEffect(() => {
     emitter.emit(WidgetEvent.WidgetExpanded, expanded);
--- a/packages/widget/src/components/Routes/RoutesExpanded.tsx
+++ b/packages/widget/src/components/Routes/RoutesExpanded.tsx
@@ -114,7 +114,7 @@ export const RoutesExpandedElement = () => {
       !isFetching &&
       isFetched &&
       emitter.emit(WidgetEvent.RoutesAvailable, routesAvailable);
-  }, [routesAvailable, emitter, isLoading, isFetching, isFetched, routes]);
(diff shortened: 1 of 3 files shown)


packages/widget/src/components/Routes/RoutesExpanded.tsx
```diff
@@ -105,10 +105,17 @@ export const RoutesExpandedElement = () => {
...
   const toAddressUnsatisfied = currentRoute && requiredToAddress && !toAddress;
   const allowInteraction = account.isConnected && !toAddressUnsatisfied;
 
+  useEffect(() => {
+    !isLoading &&
+      !isFetching &&
+      isFetched &&
+      emitter.emit(WidgetEvent.RoutesAvailable, routesAvailable);
+  }, [routesAvailable, emitter, isLoading, isFetching, isFetched]);
+
```
chybisov, line 118: As we discussed in Discord, it would make more sense to have it fire from the `useRoutes` hooks inside the query. That way we can subscribe to the event no matter which widget variant we use - compact or wide. Also, I'd return the routes array so integrators can also make use of it.

packages/widget/src/components/Routes/RoutesExpanded.tsx
```diff
@@ -105,10 +105,17 @@ export const RoutesExpandedElement = () => {
...
-
+  const routesAvailable = (routes && routes?.length > 0) || false;
   const toAddressUnsatisfied = currentRoute && requiredToAddress && !toAddress;
   const allowInteraction = account.isConnected && !toAddressUnsatisfied;
 
+  useEffect(() => {
+    !isLoading &&
+      !isFetching &&
+      isFetched &&
+      emitter.emit(WidgetEvent.RoutesAvailable, routesAvailable);
```
chybisov, line 116: Please use `if` here :)

packages/widget/src/types/events.ts
```diff
@@ -6,6 +6,7 @@ export enum WidgetEvent {
   RouteExecutionCompleted = 'routeExecutionCompleted',
   RouteExecutionFailed = 'routeExecutionFailed',
   RouteHighValueLoss = 'routeHighValueLoss',
+  RoutesAvailable = 'routesAvailable',
```
chybisov, line 9: If we are going to return the routes array, maybe it would be better to call it AvailableRoutes. In that case, we can assume that there can be no available routes, while with RoutesAvailable it looks like we always return only available ones 😄

chybisov (changes requested):

packages/widget/src/components/Routes/RoutesExpanded.tsx
dennyscode: Thanks! Good points:
a.) It´s sitting within the useRoutes hook now
b.) It´s returning the routes-array

packages/widget/src/types/events.ts
dennyscode: Nice! It´s now renamed to "AvailableRoutes"

packages/widget/src/components/Routes/RoutesExpanded.tsx
dennyscode: Yup, using that if-condition now

DNR500 (approved):

PR #267, opened 2024-07-08
fix: json assertions

DNR500:
Jira: [LF-9087](https://lifi.atlassian.net/browse/LF-9087)

We were have some issues in some of the dev tools that we support related to json assertions.

- Next.js pages router would report error when trying to our ClientOnly method of importing the Widget
- Remix would also report errors in its terminal output when trying to use the WIdgetSkeleton

We had originally tried to put json assertions in place but reverted the change as it caused problems with the use of Create React App.

In regards to this, this PR..

- reintroduces the json assertions
- adds a configuration work around to the create-react-app example
- presents update `nextjs` (next v14) and `nextjs-page-router` (next v13)  examples to document best usage in Pages and App router
- Adds the WidgetSkeleton to the `remix` example

in addition..

- Also adds a fix to the vite example for the connect button
- Adds minor updates to the documentation in the examples
- fixes some minor linting issue

Though this change also helps improve the usage of the WidgetSkeleton there are still some issues relating to its usage - these might be best explored in future tickets.
- The WidgetSkeleton triggers some compilation errors with used in Next.js 13 and its App router (this doesn't happen in Next 14 or Next 13 Pages router - see notes the `nextjs-page-router` example)
- Remix the WidgetSkeleton seems to only be usable in the fallback of the suspense block, usage in other place seems to cause issues in how the widget renders afterwards - this seems to only happen with remix 

Testing
I've tested the Widget updates locally using the examples.

Release considerations

- Once the released we will need to do an additional PR to bump version numbers in the examples folders

[LF-9087]: https://lifi.atlassian.net/browse/LF-9087?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/components/Step/StepTimer.tsx
```diff
@@ -1,7 +1,8 @@
 import type { LiFiStepExtended } from '@lifi/sdk';
 import { useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
-import { useTimer } from 'react-timer-hook';
+import pkg from 'react-timer-hook';
```
DNR500, line 4: added this as there was an error from one of the examples about this package being common js

PR #268, opened 2024-07-10
feat: improve review page with route tracker, added new fee card and improved route card

chybisov:
1. Introduced a fee card displaying the minimum amount received and token rates.
2. Route cards now display token rates with subscript representation for long decimals, which can be reversed on click.
3. Added a route refresh button to the review page, and the route automatically updates every 1 minute.
4. Price impact is now shown next to the dollar amount of the destination token.
5. Amounts less than $0.01 are now shown as `<$0.01` instead of `$0.00` or `$0`.
6. The title "LI.FI Dex Aggregator via LI.FI" step tool is now displayed as "LI.FI Dex Aggregator".
7. Fees are now correctly sorted within tooltips.

image
image


DNR500 (approved):
Looks good

PR #276, opened 2024-07-19
feat: changing height and present widget better for mobile

DNR500:
This branch looks to facilitate some of the changes for following tickets which are kind of inter-related

Jira: [LF-3514](https://lifi.atlassian.net/browse/LF-3514)
Jira: [LF-8614](https://lifi.atlassian.net/browse/LF-8614)
Jira: [LF-8615](https://lifi.atlassian.net/browse/LF-8615)

New controls have been added to facilitate testing and development when working with widget height changes. In the Design section of the Playground, you will now find a new option labeled 'Layout.' This option reveals additional controls within the Playground settings. A toggle switch is available in these settings, allowing you to easily enable or disable this feature.

This toggle currently covers the 'bookmarks store' control that sits in the 'Playground settings'- this is purely to mock out data to develop against and that mock data may not work as expected in some parts of the widget. 

**Layout controls**
Layout-Design-Control

Presents controls for - default, restricted max height, restricted height, full height.

**Default**
Layout-Default

**Restricted max height**
Layout-restricted-max-height

**Restricted height**
Layout-restricted-height

**Full height**
Layout-full-height

**Playground settings**

*When layout is NOT full height
Playground-settings-when-not-full-height

*When layout is full height - controls to show a mock header and mock footer are available
Playground-settings-when-full-height

**How this functionality works with the widget config**

We have 4 conceptual ways of dealing with the widgets: **default, restricted max height, restricted height, full height.**

**Default**
This requires no change to config but fundamentally works in the same way as the restricted max height. Its present the  default max height of the widget as 682px.

**Restricted max height**
To use this layout you can add a max height property to the container section of the config. You should use a number not a string value to provide the max height in px. And its advisable to use a value above 682 which is the default value of the widget.

```
const widgetConfig: WidgetConfig = {
  ...
  theme; {
    container: {
      maxHeight: 820
    }
  }
 ...
}
```

When using restricted max height any pages inside the widget that are bigger than the max height provided will present scrolling to allow content to be reachable. 

When pages are smaller than the max height the Widgets outer container, the 'widget-app-expanded-container', should shrink to take up only the minimum space required - generally pages inside the Widget will occupy the minimum space needed.

**Restricted height**
To use this layout you can add a height property to the container section of the config. You should use a number not a string value to provide the height in px. And its advisable to use a value above 682 which is the default value of the widget.

```
const widgetConfig: WidgetConfig = {
  ...
  theme; {
    container: {
      height: 820
    }
  }
  ...
}
```

When using restricted height all pages in the Widget should, as a minimum, occupy the full height stated and any pages inside the widget that are bigger than the height provided will present scrolling to allow content to be reachable.

Widgets outer container, the 'widget-app-expanded-container', will remain fixed at the height presented in the config.

Its recommend NOT to attempt to use the containers height and maxHeight together - in the Widget they present different modes of operation.

**Full height**
Full height has initially been included to better present the widget where less screen real estate is available such as on mobile devices. It assumes that the widget will make up the majority of the content and functionality on a page and where possible will allow scrolling via the page itself. With this layout the widget will attempt to occupy the full height of the HTML element containing it.

Full height layout can feature number of different properties in the config - we will break each of these down

```
const widgetConfig: WidgetConfig = {
  ...
  variant: 'compact'
  theme; {
    container: {
      display: 'flex',
      height: '100%'
    }
    header: {
      position: 'fixed',
      top: 0,
    },
  }
  ...
}
```

- **variant** - required - full height is designed to work with the 'compact' variant - 'compact 'itself is already built to work with smaller screen spaces in mind.
- **container** - required - this should feature ` display: 'flex',` and ` height: '100%'` to tell the Widget to occupy the full space of the containing HTML element and to also allow the use of flex in some parts of the Widget to try to better use available screen space.
- **header** - optional - in other layouts such as restricted max height and restricted height the widget headers are 'sticky' so will stay in position when the user scrolls. This can be more trick to do when using the Widget in full height layout so this option can help you achieve a similar effect when using the full height option. When used this will position the widgets page header in relation to the window and remove it from the static flow of the document. 
  -  `position: 'fixed',` - sets up the header positioning
  - `'top: 0,'` - the top property should allow the header to be position allowing for things like headers or site navigation. that might sit above the widget. For example you might want to change it to `'top: 48,'`
  -  When `header` is not included in the config the Widgets page header will remain in the static flow of the document will be part of the scrolling content - so on pages that can scroll the header may scroll out view.

**Using a full height Widget on a page**
To present the widget well for something like a mobile experience the pages HTML & CSS outside of the Widget will also have to be considered as well as the widget settings. With the Widget surrendering its height to its external HTML container this needs to be handled well by the containing application.

- Viewport meta updates may be need for the page to correctly perform well as expected on mobile
  - eg `<meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=no">`
- You may also want to think about you page behaves on in terms of occupying the fully page height
  -  for example In the Widget Playground we use '100dvh' with the min-height css which means that if widgets pages are bigger than this that the page can still scroll to allow access to that widget content. Also that when widgets pages are smaller than the viewport they can scale and position elements to better use the available space to occupy the full screen height.
  - You should add `overscroll-behavior: none;` to the root/body of your page to prevent undesired scrolling behaviour
  - Placement on the page in relation to headers and footers may also have to be considered.

[LF-3514]: https://lifi.atlassian.net/browse/LF-3514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-8614]: https://lifi.atlassian.net/browse/LF-8614?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-8615]: https://lifi.atlassian.net/browse/LF-8615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/components/Routes/RoutesExpanded.tsx
```diff
@@ -43,13 +46,18 @@ export const RoutesExpanded = () => {
   const match = Boolean(element?.props?.children);
 
   return (
-    <CollapseContainer>
+    <CollapseContainer id="CollapseContainer">
```
chybisov, line 49: Let's add this id attribute here please to be consistent https://github.com/lifinance/widget/blob/main/packages/widget/src/utils/elements.ts#L1

packages/widget-playground/src/components/DrawerControls/DesignControls/LayoutControls.tsx
```diff
@@ -0,0 +1,275 @@
...
+              {inputLabel[selectedLayoutId]}
+            </label>
+            {(heightValue && heightValue < defaultMaxHeight) || !heightValue ? (
+              <CapitalizeFirstLetter variant="caption" sx={{ paddingLeft: 1 }}>
+                {`${defaultMaxHeight}px minimum`}
+              </CapitalizeFirstLetter>
+            ) : null}
+          </CardRowColumn>
+          <Input
+            id="layout-height-input"
```
chybisov, line 258: Let's add this id attribute here please to be consistent https://github.com/lifinance/widget/blob/main/packages/widget/src/utils/elements.ts#L1

packages/widget/src/components/AppContainer.tsx
```diff
@@ -64,9 +109,12 @@ export const AppContainer: React.FC<PropsWithChildren<{}>> = ({ children }) => {
         id={createElementId(ElementId.ScrollableContainer, elementId)}
         variant={variant}
         enableColorScheme
+        paddingTopAdjustment={positionFixedAdjustment}
         // ref={ref}
       >
-        <FlexContainer disableGutters>{children}</FlexContainer>
+        <FlexContainer disableGutters id="app-level-flex-container">
```
chybisov, line 115: Let's add this id attribute here please to be consistent https://github.com/lifinance/widget/blob/main/packages/widget/src/utils/elements.ts#L1

packages/widget/src/components/Routes/RoutesExpanded.tsx
```diff
@@ -112,15 +120,20 @@ export const RoutesExpandedElement = () => {
...
-        <Container enableColorScheme>
-          <ScrollableContainer>
+      <RouteNestedGrow
+        timeout={timeout.enter}
+        in={expanded}
+        mountOnEnter
+        unmountOnExit
+      >
+        <Container enableColorScheme minimumHeight={isLoading}>
+          <ScrollableContainer id="ScrollableContainer">
```
chybisov, line 136: Let's add this id attribute here please to be consistent https://github.com/lifinance/widget/blob/main/packages/widget/src/utils/elements.ts#L1

packages/widget/src/components/TokenList/TokenList.tsx
```diff
@@ -82,7 +81,11 @@ export const TokenList: FC<TokenListProps> = ({
     !tokenSearchFilter;
 
   return (
-    <Box ref={parentRef} style={{ height, overflow: 'auto' }}>
+    <Box
+      ref={parentRef}
+      style={{ height, overflow: 'auto' }}
+      id="token-list-parent"
```
chybisov, line 87: Let's add this id attribute here please to be consistent https://github.com/lifinance/widget/blob/main/packages/widget/src/utils/elements.ts#L1

packages/widget/src/components/ChainSelect/MobileChainSelect.tsx
chybisov, line 1: I remember we discussed merging this functionality with the regular ChainSelect component 👀

chybisov (changes requested):

packages/widget/src/components/Routes/RoutesExpanded.tsx
DNR500: I think can remove these - I've accidently left them in from debug the routes positioning so they aren't really needed

packages/widget/src/components/ChainSelect/MobileChainSelect.tsx
DNR500: Have deleted the MobileChainSelect code

packages/widget/src/components/TokenList/TokenList.tsx
DNR500: I think can remove these - I've accidently left them in from debug the routes positioning so they aren't really needed

packages/widget/src/components/Routes/RoutesExpanded.tsx
DNR500: I think can remove these - I've accidently left them in from debug the routes positioning so they aren't really needed

packages/widget/src/components/AppContainer.tsx
DNR500: I think can remove these - I've accidently left them in from debug the routes positioning so they aren't really needed

packages/widget-playground/src/components/DrawerControls/DesignControls/LayoutControls.tsx
DNR500: I think can remove these - I've accidently left them in from debug the routes positioning so they aren't really needed

chybisov (approved):

PR #292, opened 2024-09-02
feat: add explorer link to the support card

DNR500:
Jira: [LF-9382](https://lifi.atlassian.net/browse/LF-9382)

Screenshot 2024-09-03 at 12 16 19

This changes the name of the support id card to `Transfer Id` and adds a link to open a new tab from the card that appears at the end of a transactions on the transaction details page.

It defaults to using LiFi scan - `https://scan.li.fi` and will open the page

https://scan.li.fi/tx/0x48018bad1907c531fabf0efa509eebf25ee4f0f37c916732648b912c53a7a223

The link is configurable via widget config, for example the addition of the following

```
const config = {
  ...
  explorerUrl: 'https://jumper.exchange/scan',
  ...
}
```
Will open the page 

https://jumper.exchange/scan/tx/0x48018bad1907c531fabf0efa509eebf25ee4f0f37c916732648b912c53a7a223/

In the config only the base url is needed `/tx/{hash}` is added internally by the widget

[LF-9382]: https://lifi.atlassian.net/browse/LF-9382?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/widget-playground/src/defaultWidgetConfig.ts
+++ b/packages/widget-playground/src/defaultWidgetConfig.ts
@@ -206,7 +206,6 @@ export const widgetBaseConfig: WidgetConfig = {
 export const defaultWidgetConfig: Partial<WidgetConfig> = {
   ...widgetBaseConfig,
   appearance: 'auto',
-  explorerUrl: 'https://scan.li.fi/tx/',
   theme: {
     palette: {
       primary: {
--- a/packages/widget-playground/src/defaultWidgetConfig.ts
+++ b/packages/widget-playground/src/defaultWidgetConfig.ts
@@ -206,6 +206,7 @@ export const widgetBaseConfig: WidgetConfig = {
 export const defaultWidgetConfig: Partial<WidgetConfig> = {
   ...widgetBaseConfig,
   appearance: 'auto',
+  explorerUrl: 'https://scan.li.fi/tx/',
   theme: {
     palette: {
       primary: {
--- a/packages/widget/src/pages/TransactionDetailsPage/SupportIdCard.tsx
+++ b/packages/widget/src/pages/TransactionDetailsPage/SupportIdCard.tsx
@@ -6,6 +6,8 @@ import { CardIconButton } from '../../components/Card/CardIconButton.js';
 import { CardTitle } from '../../components/Card/CardTitle.js';
 import { useWidgetConfig } from '../../providers/WidgetProvider/WidgetProvider.js';
 
+const lifiExplorerUrl = 'https://scan.li.fi';
+
 interface SupportIdCardProps {
   supportId: string;
 }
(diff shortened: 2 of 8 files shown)


packages/widget-playground/src/defaultWidgetConfig.ts
```diff
@@ -206,6 +206,7 @@ export const widgetBaseConfig: WidgetConfig = {
 export const defaultWidgetConfig: Partial<WidgetConfig> = {
   ...widgetBaseConfig,
   appearance: 'auto',
+  explorerUrl: 'https://scan.li.fi/tx/',
```
DNR500, line 209: using a string here to preset the url prefix for the explorer. We could potentially use a function in place here if urls were to also feature a suffix. eg

```
exploreUrl: (tx: string) => `http://prefix/${tx}/suffix`,
```
But for the use case for l LiFi and Jumper scan I don't think that is needed

packages/widget/src/pages/TransactionDetailsPage/SupportIdCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+            gap: 1,
+            display: 'flex',
+            marginRight: 2,
+            marginTop: 1,
+          }}
+        >
+          <CardIconButton size="small" onClick={copySupportId}>
+            <ContentCopyRounded fontSize="inherit" />
+          </CardIconButton>
+          {explorerUrl ? (
```
DNR500, line 52: Currently the button for the link will only appear if an `exploreUrl` is featured in the config - there is a question here as to whether we would want `explorerUrl` to default to `https://scan.li.fi/tx/` if explorerUrl is not added to the Widget config?

chybisov: Yes please, let's always fallback to li.fi scan

packages/widget-playground/src/defaultWidgetConfig.ts
chybisov: @DNR500 I guess we can provide just the basic url in options like https://scan.li.fi and the rest should follow conventions (`/tx/{hash}`).

DNR500: Makes sense

packages/widget/src/pages/TransactionDetailsPage/SupportIdCard.tsx
```diff
@@ -0,0 +1,71 @@
+import { ContentCopyRounded, OpenInNew } from '@mui/icons-material';
+import { Box, Typography } from '@mui/material';
+import { useTranslation } from 'react-i18next';
+import { Card } from '../../components/Card/Card.js';
+import { CardIconButton } from '../../components/Card/CardIconButton.js';
+import { CardTitle } from '../../components/Card/CardTitle.js';
+import { useWidgetConfig } from '../../providers/WidgetProvider/WidgetProvider.js';
+
+const lifiExplorerUrl = 'https://scan.li.fi';
```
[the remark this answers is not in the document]
DNR500: Already on it :-)

chybisov (approved):

PR #293, opened 2024-09-04
feat: allow configuration of explorer links

DNR500:
JIra: [LF-9882](https://lifi.atlassian.net/browse/LF-9882)

Allows configuration of explorer urls in the Widget

[LF-9882]: https://lifi.atlassian.net/browse/LF-9882?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/widget/src/hooks/useExplorer.ts
+++ b/packages/widget/src/hooks/useExplorer.ts
@@ -17,7 +17,10 @@ export const useExplorer = () => {
     return sanitiseBaseUrl(baseUrl);
   };
 
-  const getTransactionLink = (txHash: string, chain?: Chain) => {
+  const resolveChain = (chain: Chain | number) =>
+    Number.isFinite(chain) ? getChainById(chain as number) : (chain as Chain);
+
+  const getTransactionLink = (txHash: string, chain?: Chain | number) => {
     if (!chain) {
       const baseUrl = explorerUrls?.internal?.[0]
         ? sanitiseBaseUrl(explorerUrls?.internal[0])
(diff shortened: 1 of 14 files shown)


packages/widget/src/hooks/useExplorer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  const getAddressLinkByChainId = (address: string, chainId: number) => {
+    return getAddressLink(address, getChainById(chainId));
+  };
+
+  return {
+    getTransactionLink,
+    getTransactionLinkByChainId,
+    getAddressLink,
+    getAddressLinkByChainId,
```
DNR500, line 53: In some places we already have access to the chain object and in other places with only have access to the chain id so have some additional functions to deal with this

packages/widget/src/hooks/useExplorer.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    return `${getBaseUrl(chain)}/tx/${txHash}`;
+  };
+
+  const getTransactionLinkByChainId = (txHash: string, chainId: number) => {
+    return getTransactionLink(txHash, getChainById(chainId));
+  };
+
+  const getAddressLink = (address: string, chain?: Chain) => {
+    // TODO: consider here that LiFi explorer supports /wallet/ rather than /address/
```
DNR500, line 36: Lifi Explorer (and Jumper Scan too) `/wallet/` gets used in place of  `/address/`

https://github.com/lifinance/lifi-explorer/blob/develop/packages/explorer/src/config/routes.ts

In the current WIP version I have set this up so you can see the problem.

1. Connect your wallet
2. Select Arbitrum as your chain in the wallet
3. And then click the explorer link from the widget wallet menu

Should we change the widget configuration on this to allow the tx and address paths to be overridable?

something like this?

```
explorerUrls: {
    42161: {
      baseUrl: 'https://jumper.exchange/scan',
    },
    internal: {
      baseUrl: 'https://jumper.exchange/scan',
      addressPath: '/wallet/',
      txPath: '/transaction/',
   },
}
```

packages/widget/src/types/widget.ts
```diff
@@ -225,6 +225,7 @@ export interface WidgetConfig {
   languages?: WidgetLanguages;
   languageResources?: LanguageResources;
   explorerUrl?: string;
```
chybisov, line 227: Let's remove this :)

packages/widget/src/hooks/useExplorer.ts
DNR500: remove the ..ByChainId functions and consolidated that functionality into the other functions

packages/widget/src/hooks/useExplorer.ts
DNR500: This is a possible option in future should we need it. Staying with the current structure for now

chybisov (approved):

PR #294, opened 2024-09-04
fix: reactive chain and token properties from config

DNR500:
Jira: [LF-9902](https://lifi.atlassian.net/browse/LF-9902)

### **How to set up testing in the playground**
Some tooling has been developed to test this functionality in the playground.

To access it
- Open the Playground
- Go to the “Playground settings”
- Click the Dev view button
- A control for “Form values” will appear in the Design tools
- You can use this to dynamically change the values 

There are two ways you can update the form values in the Widget
- using config - this uses reactive values in the config and requires some management of those values for updates
- using the formRef - this provides an function that you can use to update values in the widgets form

**Note: Config overrides URL params**
Important point from this PR - Config values override URL parameters.

So you need to ensure that on first page load non of the form values that can be set in the URL are featured in the config that the Widget initialises.

So `fromAmount`, `fromChain`, `fromToken`, `toAddress`, `toChain`, `toToken` and `toAmount` should not be set to allow the URL to perform the initial set up.

On first page load if you have both values in the config and the url then the url search params will be rewritten to match the config values

### **Updating form values via the config**
Once widget has initialised and has full rendered  you can update the form values in the Widget by updating the config.

To perform an update you should only include the form values in the config that you want to change and ensure this is passed to the Widget.

So if you want to change the fromChain and fromToken and nothing else you should include only include those values 

You also need to set the formUpdateKey. This should be a unique randomly generated key. Using this key ensures that the form values are updated in the widget - essentially forcing an update in the Widget even if the config is otherwise unchanged. This can avoid some edge case issue that might arise when setting values in the Widget via a mix of config and user actions via the Widgets UI.

So your new config might look like this 

```
{
    fromChain: 10,
    fromToken: ‘0x94b008aA00579c1307B0EF2c499aD98a8ce58e58’,
    // use the date object to generate a unique value
    formUpdateKey: new Date().valueOf().toString()

    // might feature other config values but should not include other form values…
}
```

You can also reset the form values and their fields to an empty state using undefined

This example resets only the fromChain and fromToken fields
```
{
    fromChain: undefined,
    fromToken: undefined,
    // use the date object to generate a unique value
    formUpdateKey: new Date().valueOf().toString()

    // might feature other config values but should not include other form values…
}
```

Note that `undefined` as a value is used to make the Widget reset to an empty form field for that property. The absence property rom the config object means they will remain unchanged.

### **Updating form values via the form ref**

This method provides developers using the Widget a way to set the form values directly in Widget using a function call. This approach is more imperative and might be better suited to some situations. By passing a ref object to the widget you can access a function to set values directly in the form.

For example

```
import type { FormState } from '@lifi/widget';
import { LiFiWidget } from '@lifi/widget';

const formRef = useRef<FormState>(null);

<LiFiWidget
  config={config}
  integrator="li.fi-playground"
  formRef={formRef}
/>

```

Once initialised you can use the ref as follows

```
formRef.current?.setFieldValue( ‘fromChain’, 10, { setUrlSearchParam: true });

```
 
I think this currently update the url and the widget fields as expected.

**To set fromChain and fromToken**
```
formRef.current?.setFieldValue( 
    'fromChain',
    10, 
    { setUrlSearchParam: true }
);
formRef.current?.setFieldValue(
    'fromToken', 
    '0x94b008aA00579c1307B0EF2c499aD98a8ce58e58', 
    { setUrlSearchParam: true }
);
```
To reset fromChain and fromToken
```
formRef.current?.setFieldValue( 
    'fromChain',
    undefined, 
    { setUrlSearchParam: true }
);
formRef.current?.setFieldValue(
    'fromToken', 
    undefined 
    { setUrlSearchParam: true }
);
```

**To set fromAmount**
```
formRef.current?.setFieldValue( 
   'fromAmount',
    '10', 
    { setUrlSearchParam: true }
);
```
to reset fromAmount
```
formRef.current?.setFieldValue(
    'fromAmount', 
    undefined 
    { setUrlSearchParam: true }
);
```

**To set toAddress**
```
formRef.current?.setFieldValue( 
    ‘toAddress’,
    {
      address: '0x4577a46A3eCf44E0ed44410B7793977ffbe22CE0',
      chainType: ChainType.EVM,
    }, 
    { setUrlSearchParam: true }
);
```
You can also include a name to display
```
formRef.current?.setFieldValue( 
    'toAddress',
    {
      name: 'Lenny',
      address: '0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea9',
      chainType: ChainType.EVM,
    }, 
    { setUrlSearchParam: true }
);
```
Or you can just set the address as a string
```
formRef.current?.setFieldValue( 
    'toAddress',
    '0x552008c0f6870c2f77e5cC1d2eb9bdff03e30Ea9'
    { setUrlSearchParam: true }
);
```
to reset toAddress
```
formRef.current?.setFieldValue(
    'toAddress', 
    undefined 
    { setUrlSearchParam: true }
);
```

[LF-9902]: https://lifi.atlassian.net/browse/LF-9902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/providers/WidgetProvider/WidgetProvider.tsx
```diff
@@ -29,45 +27,9 @@ export const WidgetProvider: React.FC<
...
-      // Get search params from URL
-      const searchParams = Object.fromEntries(
-        new URLSearchParams(window?.location.search),
-      );
-      // Prevent using fromToken/toToken params if chain is not selected
-      ['from', 'to'].forEach((key) => {
-        if (searchParams[`${key}Token`] && !searchParams[`${key}Chain`]) {
-          delete searchParams[`${key}Token`];
-        }
-      });
```
DNR500, line 41: The logic here has been moved to the URLSearchParamsBuilder and now rather than overridding the config values is used to set form values in the form store

https://github.com/lifinance/widget/blob/fbbd18c4a1e846e0570417b67b12ac47157418bb/packages/widget/src/stores/form/URLSearchParamsBuilder.tsx#L59-L68

and 

https://github.com/lifinance/widget/blob/fbbd18c4a1e846e0570417b67b12ac47157418bb/packages/widget/src/stores/form/URLSearchParamsBuilder.tsx#L19-L45

packages/widget/src/stores/form/FormStore.tsx
```diff
@@ -10,44 +10,28 @@ import type { FormStoreStore } from './types.js';
...
-    [
-      fromAmount,
-      fromChain,
-      fromToken,
-      hiddenToAddress,
-      toAddress,
-      toChain,
-      toToken,
-    ],
+    [config, hiddenToAddress],
```
DNR500, line 34: There is a case that required this code to be changed. 

Before this change the issue could be recreated by doing the following, 
- In the playground if the user selected a from chain and token option from the design tools
- The user than changes the from chain and token in the widget itself to something else
- The user than tries to click the same from chain and token in the design tools as they did previously
- This last change would fail to update the from chain and token in the widget

I've change this code to look for general changes of the config object - so an immutable approach will allow those values to be updated. It might mean that this code is called more frequently though and we should discuss this approach.

packages/widget/src/stores/form/FormUpdater.tsx
```diff
@@ -50,21 +57,42 @@ export const FormUpdater: React.FC<{
...
+    // set values from config
+    if (defaultValues.toAddress) {
+      initialiseSendToWallet(true);
+    }
+
+    // TODO: consider using null in the config to specify when a value should be set to its default
+    //  look at using formDefaultValues for this
+    setUserAndDefaultValues(
+      removeEmptyValuesProperties(defaultValues, account.chainId),
+    );
```
DNR500, line 69: This was previously using a ref to hold previous default values but risked missing comparison between config set values and user set values via the widgets UI (form state values).

The checks for this now happen against the current user values in the form state

https://github.com/lifinance/widget/blob/fbbd18c4a1e846e0570417b67b12ac47157418bb/packages/widget/src/stores/form/createFormStore.ts#L97-L106

Doing things this way, irrespective of it the form value was last set by user interaction on the widget or via the config, the last attempt to set the value should be the set value in the widget form.

packages/widget/src/components/SendToWallet/SendToWalletButton.tsx
```diff
@@ -44,8 +44,7 @@ export const SendToWalletButton: React.FC<CardProps> = (props) => {
   const hiddenToAddress = hiddenUI?.includes(HiddenUI.ToAddress);
 
   const showInstantly =
-    Boolean(!showSendToWalletDirty && toAddress && !hiddenToAddress) ||
-    requiredToAddress;
+    Boolean(toAddressFieldValue && !hiddenToAddress) || requiredToAddress;
```
DNR500, line 47: Some of the values in the Send To Wallet Button (and the Expand Button) where using form values from the config. Switching these to use form values in the state keeps the Send to Wallet button and its open and closed state working as expect

packages/widget/src/components/SendToWallet/SendToWalletButton.tsx
```diff
@@ -103,14 +98,23 @@ export const SendToWalletButton: React.FC<CardProps> = (props) => {
...
-    ? showSendToWallet || (showInstantly && !showSendToWalletDirty)
-    : showSendToWallet || showInstantly;
+  // The collapse opens instantly on first page load/component mount when there is an address to display
+  // After which it needs an animated transition for open and closing.
+  // collapseTransitionTime is used specify the tranistion time for opening and closing
+  const [collapseTransitionTime, setCollapseTransitionTime] = useState(0);
+
+  useEffect(() => {
+    // this should fire after the collapse have been opened instantly
+    setCollapseTransitionTime(225);
```
chybisov, line 108: Do you think we could change this to use useRef to avoid rerender? Not a fan of updating state after mount if effect 😅🙏

packages/widget/src/stores/form/FormStore.tsx
```diff
@@ -47,17 +100,84 @@ export const FormStoreProvider: React.FC<PropsWithChildren> = ({
...
+            setSelectedBookmark(value);
+          }
+
+          return address;
+        },
+      };
+
+      formApiRef.current = {
+        setFieldValue: (fieldName, value, options) => {
+          const santisedValue = (
```
chybisov, line 161: Small typo in `sanitized`

packages/widget/src/stores/form/FormStore.tsx
```diff
@@ -1,15 +1,42 @@
...
+    : toAddress?.address || formDefaultValues.toAddress,
+});
+
+interface FormStoreProviderProps extends PropsWithChildren {
+  formRef?: FormRef;
+}
+
+export const FormStoreProvider: React.FC<FormStoreProviderProps> = ({
   children,
+  formRef,
```
chybisov, line 36: Maybe we should wrap this in `forwardRef` instead of passing formRef as a prop?

packages/widget/src/stores/form/useFormRef.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { FormRef } from '../../types/widget.js';
+import { HiddenUI } from '../../types/widget.js';
+import type { FormValuesState, GenericFormValue } from './types.js';
+
+export const useFormRef = (formStore: FormValuesState, formRef?: FormRef) => {
+  const { setSendToWallet } = useSendToWalletStore();
+  const { setSelectedBookmark } = useBookmarkActions();
+  const { hiddenUI } = useWidgetConfig();
+
+  const hiddenToAddress = hiddenUI?.includes(HiddenUI.ToAddress);
```
chybisov, line 15: Let's move this inside `sanitizeValue` function closer to its usage.

packages/widget/src/stores/form/useFormRef.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const santizeValue: {
```
chybisov, line 17: Typo santizeValue -> san**i**tizeValue

chybisov, line 17: Do you think we can move this function inside useImperativeHandle function?

packages/widget/src/stores/form/useFormRef.ts
```diff
@@ -0,0 +1,66 @@
...
+import { useWidgetConfig } from '../../providers/WidgetProvider/WidgetProvider.js';
+import { useBookmarkActions } from '../../stores/bookmarks/useBookmarkActions.js';
+import { formDefaultValues } from '../../stores/form/createFormStore.js';
+import { useSendToWalletStore } from '../../stores/settings/useSendToWalletStore.js';
+import type { FormRef } from '../../types/widget.js';
+import { HiddenUI } from '../../types/widget.js';
+import type { FormValuesState, GenericFormValue } from './types.js';
+
+export const useFormRef = (formStore: FormValuesState, formRef?: FormRef) => {
+  const { setSendToWallet } = useSendToWalletStore();
```
chybisov, line 11: Can we select only `setSendToWallet` from the store to avoid re-renders when other store values change?

packages/widget/src/types/widget.ts
```diff
@@ -228,7 +234,46 @@ export interface WidgetConfig {
...
+
+export type FormFunctions =
+  | {
+      setFieldValue: SetFieldValueFunc;
+    }
+  | undefined;
+
+export type FormRef = MutableRefObject<FormFunctions>;
+
+export interface FormRefProp {
```
chybisov, line 272: Let's be consistent: FormRefProp -> FormRefProp**s**

packages/widget/src/types/widget.ts
```diff
@@ -228,7 +234,46 @@ export interface WidgetConfig {
...
+  key: K,
+  value: FieldValues[K],
+  options?: FormFieldOptions,
+) => void;
+
+export type FormFunctions =
+  | {
+      setFieldValue: SetFieldValueFunc;
+    }
+  | undefined;
```
chybisov, line 268: Why this can be undefined?

packages/widget/src/types/widget.ts
```diff
@@ -228,7 +234,46 @@ export interface WidgetConfig {
...
+export type FieldValues = {
+  fromChain: number | undefined;
+  fromToken: string | undefined;
+  toChain: number | undefined;
+  toToken: string | undefined;
+  fromAmount: number | string | undefined;
+  toAddress: ToAddress | string | undefined;
+};
+
+export type SetFieldValueFunc = <K extends FieldNames>(
```
chybisov, line 258: Let's make it `SetFieldValueFunction` to align with `FormFunctions`

packages/widget/src/types/widget.ts
DNR500: Because the initial state of the useRef will be undefined before the functions are added to it

packages/widget/src/stores/form/FormStore.tsx
```diff
@@ -47,17 +93,34 @@ export const FormStoreProvider: React.FC<PropsWithChildren> = ({
...
+      intialiseDefaultValues(
+        reactiveFormValues,
+        fromAmount,
+        toAddress,
+        hiddenToAddress,
+      ),
+    );
   }
 
+  useFormRef(storeRef.current.getState(), formRef);
```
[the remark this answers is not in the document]
DNR500: I intially had formRef as the first option but eslint complains about it being options complains about potentially being. If formRef isn't set its undefined

I can pass storeRef.current instead

packages/widget/src/stores/form/FormStore.tsx
DNR500: I think that only applies i you want to use 'ref' as the prop name - I think keeping this as formRef is actually a good thing here. I always think that ref is generally used to refer to DOM elements as it most common use. Also come React 19 we are going to have to stop using forwardRef anyways

packages/widget/src/stores/form/useFormRef.ts
DNR500: See what you mean. changing that to useSendToWalletActions rather than useSendToWalletStore

packages/widget/src/stores/form/FormStore.tsx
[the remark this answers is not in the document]
DNR500: I've made the change to pass storeRef.current instead, I've kept the formRef as the second argument to keep eslint happy

packages/widget/src/stores/form/FormStore.tsx
chybisov: Yeah, in React 19 we will have ref prop without the need to wrap the component in forwardRef, however to use ref prop now it is still required 😄 Agree, ref is often used for DOM elements and making a component controlled with ref is also a common use-case and we use `useImperativeHandle` for that.

packages/widget/src/types/widget.ts
DNR500: I'll change this to

```
export type FormFunctions = {
  setFieldValue: SetFieldValueFunction;
};
```

And point of use should change to

```
const formRef = useRef<FormFunctions | undefined>()
```

That seems cleaner

chybisov (approved):

PR #304, opened 2024-09-20
feat: add emitter event on location change

dennyscode:
Ticket: https://lifi.atlassian.net/browse/LF-10006


packages/widget/src/types/events.ts
```diff
@@ -14,6 +14,7 @@ export enum WidgetEvent {
   ReviewTransactionPageEntered = 'reviewTransactionPageEntered',
   WalletConnected = 'walletConnected',
   WidgetExpanded = 'widgetExpanded',
+  Page = 'page',
```
chybisov, line 17: Let's name it `PageEntered`

packages/widget/src/types/events.ts
```diff
@@ -14,6 +14,7 @@ export enum WidgetEvent {
   ReviewTransactionPageEntered = 'reviewTransactionPageEntered',
```
chybisov, line 14: We should deprecate `ReviewTransactionPageEntered`

packages/widget/src/types/events.ts
```diff
@@ -30,6 +31,7 @@ export type WidgetEvents = {
   reviewTransactionPageEntered?: Route;
   walletConnected: WalletConnected;
   widgetExpanded: boolean;
+  page: string;
```
chybisov, line 34: Let's create a union type from `navigationRoutes` constant instead of string that lists all relevant pages we would like to track, except maybe `home`? Let me please know what do you think :)

packages/widget/src/AppRoutes.tsx
```diff
@@ -119,6 +122,12 @@ const routes: RouteObject[] = [
 ];
 
 export const AppRoutes = () => {
+  const emitter = useWidgetEvents();
   const element = useRoutes(routes);
+  const location = useLocation();
+
+  useEffect(() => {
+    emitter.emit(WidgetEvent.Page, location.pathname);
+  }, [emitter, location.pathname]);
```
chybisov, line 131: Instead of making this component (the whole widget?) to re-render every time we navigate to a new page, let create a wrapper around `useNavigate` hook and emit event on user action. Basically, we can create a custom `useNavigate` and use react-router one inside together with an even emitter.

chybisov (changes requested):

packages/widget/src/AppRoutes.tsx
dennyscode: Nice idea!

packages/widget/src/types/events.ts
dennyscode: Good idea! I have re-used the existing `NavigationRouteType`.

Just not sure with this: To silence ts, I have used it that way here:
image

PR #305, opened 2024-09-24
Feat Add search input to chain, bridges and exchanges pages

DNR500:
JIra: [LF-8523](https://lifi.atlassian.net/browse/LF-8523)

This PR adds stick search inputs to the chains, bridges and exchanges pages

Screenshot 2024-09-24 at 20 06 58
Screenshot 2024-09-24 at 20 07 12

It reuses and makes a component out of the search input thats already in use for token selection and creates a set of reusable components across the pages.

There were a couple of caveats around the different layout height modes

- When the layout is in default or max height mode the page could potentially collapse whilst a users was searching and this looked odd so the `useFullPageInMaxHeightLayout` hook was made with the intention ensuring the page takes up the maximum amount of space available.
- Sticky mode when the Widget is in Full height layout mode requires some thought to ensure that it displays correctly. It also needs access to the header height, which it does using the `useHeaderHeight` hook.
  - Whilst working on this I noticed that `useHeaderHeight` wasn't working properly I've created an alternative implementation here that use MutationObserver as a potential work around. I've some other ideas for potentially improvements in relation to this that I'll mention on the PR too.

    - Screenshot 2024-09-24 at 19 42 46

There needed to be some thought around the checkbox functionality as well on the Bridges and Exchanges pages - mainly around the select all and deselect all checkbox. It would be good to check that if the behaviour that I have put together here is desirable.

checkbox-capture

[LF-8523]: https://lifi.atlassian.net/browse/LF-8523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/hooks/useHeaderHeight.ts
```diff
@@ -1,25 +1,40 @@
...
-      ? maxHeaderHeightSubvariantSplit
-      : hiddenUI?.includes('walletMenu') || hasExternalProvider
-        ? minHeaderHeight
-        : maxHeaderHeight;
+    return () => {
+      if (resizeObserver) {
+        resizeObserver.disconnect();
+      }
+    };
+  }, [elementId, headerHeight, headerElement, setHeaderHeight]);
```
DNR500, line 37: Using the MutationObserver here in place of the imperative if checks we were doing previously - that was getting more difficult to get to get it to work in certain situations.

For the split subvariant there was still this problem in full height layout mode. And this problem can happen on other pages.

Screenshot 2024-09-24 at 19 42 46

One improvement I think that would still be good here is to move the `useLayoutEffect` to the Header component itself and have that set a value on the Header Store - the useHeaderHeight hook would then become an access point for the height value in the store and using zustand shallow checks would actually be able to limit potential renders in the Widget where the height value it used

chybisov: @DNR500 are we sure we want to see `headerHeight` in the deps array?

packages/widget/src/hooks/useFullPageInMaxHeightLayout.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useDefaultElementId } from './useDefaultElementId.js';
+
+// When the Widgets layout is set to use max height or in its default height mode the default behaviour
+// is for pages to use only the minimum height that they need to show their content.
+// On some pages you might want to always ensure that the page always up the maximum height available.
+// To do this just add this hook at the page level.
+// The other height layout will by default always occupy the maximum space available.
+// NOTE: this hook is implicitly tied to the widget height functionality in the
+//   AppExpandedContainer, RelativeContainer and CssBaselineContainer components as defined in AppContainer.ts
+//   CSS changes in those components can have implications for the functionality in this hook
```
chybisov, line 14: This thing becomes so complex 🥲

packages/widget/src/components/Search/SearchInput.style.tsx
```diff
@@ -0,0 +1,44 @@
...
+  paddingRight: theme.spacing(1.5),
+}));
+
+interface SearchStickyContainerProps {
+  headerHeight: number;
+}
+
+export const searchContainerHeight = 64;
+
+// When the widget is in Full Height mode the
```
chybisov, line 14: Something is missing at the end of the comment?

packages/widget/src/components/Search/SearchInput.style.tsx
```diff
@@ -0,0 +1,44 @@
...
+        position: 'fixed',
+        minWidth: theme.breakpoints.values.xs,
+        maxWidth: theme.breakpoints.values.sm,
+        width: '100%',
+      }
+    : {}),
+}));
+
+// When in full height mode the list needs make an addition to the paddingTop of the SearchStickyContainer height
+// this compensations for SearchStickyContainer being fixed position and compensates for it no long being in the document flow
```
chybisov, line 37: It is not clear to me, especially the second part 😅

packages/widget/src/hooks/useHeaderHeight.ts
chybisov: > One improvement I think that would still be good here is to move the useLayoutEffect to the Header component itself and have that set a value on the Header Store - the useHeaderHeight hook would then become an access point for the height value in the store and using zustand shallow checks would actually be able to limit potential renders in the Widget where the height value it used

Should we do it now? 👀

packages/widget/src/stores/settings/useSettingsStore.ts
```diff
@@ -99,17 +99,38 @@ export const useSettingsStore = createWithEqualityFn<SettingsState>(
...
+            }
+          });
+
           return {
-            [`_enabled${toolType}`]: enabledTools,
-            [`enabled${toolType}`]: enableAll ? Object.keys(enabledTools) : [],
-            [`disabled${toolType}`]: enableAll ? [] : Object.keys(enabledTools),
+            [`_enabled${toolType}`]: updatedTools,
+            [`enabled${toolType}`]: enableKeys,
+            [`disabled${toolType}`]: disabledKeys,
```
chybisov, line 133: Due to these changes here please double-check just in case that we send the correct filtering options to the API.

packages/widget/src/hooks/useHeaderHeight.ts
DNR500: I'm doing this move now and the deps array is much simpler and doesn't need the headerHeight.

packages/widget/src/components/Search/SearchInput.style.tsx
DNR500: Have reword this

packages/widget/src/components/Search/SearchInput.style.tsx
DNR500: Have reworded this, hopefully it makes more sense

packages/widget/src/stores/settings/useSettingsStore.ts
DNR500: WIll check this in the morning

packages/widget/src/hooks/useFullPageInMaxHeightLayout.ts
DNR500: I've decided to go with a way to let CSS carry the load here.

Have deleted this hook and created a full page container component

https://github.com/lifinance/widget/blob/ddbb8410419a178363aec8fc10f6a1248f9d73b1/packages/widget/src/components/FullPageContainer.tsx#L9-L14

The css for this lives in the AppContainer

https://github.com/lifinance/widget/blob/ddbb8410419a178363aec8fc10f6a1248f9d73b1/packages/widget/src/components/AppContainer.tsx#L93-L97

And just means on the pages that we swap out the PageContainer for the FullPageContainer

https://github.com/lifinance/widget/blob/ddbb8410419a178363aec8fc10f6a1248f9d73b1/packages/widget/src/pages/SelectChainPage/SelectChainPage.tsx#L69

This method was already being used for the SelectTokenPage by setting an id on the TokenList, I've just repurposed it here.

DNR500: In relation to the layout/height code I think there are still something that we can do to simplify things further in future.

- Consolidate the logic and styles in the AppContainer.tsx files AppExpandedContainer, RelativeContainer and CssBaselineContainerProps. It does get confusing where the height should be set to get the desired result for layouts
- useTokenListHeight hook could potentially be replaced with a css implementation - I've managed to create a list in a side project that  resizes in the way that the token list in our app does using just css. Though I have had a failed attempt at getting this working already I do think that at some point its worth another attempt.

DNR500:
I've also added some fixes to make the WIdgetSkeleton look better in relation to the different layout modes

packages/widget/src/stores/settings/useSettingsStore.ts
DNR500: Have check that the bridge and exchanges are passed correctly to the options on request - it looks like its working

packages/widget/src/components/TokenList/TokenList.tsx
```diff
@@ -84,11 +81,7 @@ export const TokenList: FC<TokenListProps> = ({
     !tokenSearchFilter;
 
   return (
-    <Box
-      ref={parentRef}
-      style={{ height, overflow: 'auto' }}
-      id={createElementId(ElementId.TokenList, elementId)}
```
DNR500, line 90: This isn't needed any more as the SelectTokenPage used the FullPageContainer

packages/widget/src/components/Skeleton/WidgetSkeleton.style.tsx
```diff
@@ -57,7 +57,27 @@ export const SkeletonSendToWalletButton = styled(Button)({
...
   display: 'flex',
+  flexGrow: 1,
   justifyContent: 'flex-end',
+  alignItems: 'flex-end',
+  paddingBottom: theme.spacing(2),
+  paddingLeft: theme.spacing(3),
+  paddingRight: theme.spacing(3),
+}));
+
+export const SkeletonHeaderContainer = styled(Box)(({ theme }) => {
```
DNR500, line 70: Some widget skeleton changes were needed so that the Skeleton doesn't apear broken in the various layout modes

PR #309, opened 2024-10-08
feat: widget events for send to wallet

DNR500:
Jira: [LF-8931](https://lifi.atlassian.net/browse/LF-8931)

This PR adds a "FormFieldChanged" to the widget. This event will fire when..
- The user updates the any form values via the Widget UI - the event will only be emitted after the Widget has done its validation
- A form value is updated via a external method to the Widget UI - either via config, url search params or formRef. These do not undergo the same validation as user entered values via the UI. Strict validation of values added to things like the config should be done before adding them to the config itself.

Note: that with this implementation the initialise value from the widget config will not emit an event. 

To catch the event specifically for the send to wallet address change you will need to check the fieldName value in on the payload
```
widgetEvents.on(WidgetEvent.FormFieldChanged, (payload) => {
   if(payload.fieldName === "toAddress") {
      // sent to wallet address was changed
   }
)
```

The event payload will looks like this

```
{
  fieldName: "toAddress",
  oldValue: "0x8131aa556b78e6D25F53DA0F01c2f046AA1012d1",
  value: "0x29DaCdF7cCaDf4eE67c923b4C22255A4B2494eD7"
}

```

**Widget Events Tooling**

I've also added some basic tooling to the the playground to assist with testing and working with events. This should help with the other events we need to add to widget as well. This sits behind the dev view toogle.

To view the tool
- Open Playground, go to the Playground settings and turn on the Dev view option
- The Widget events control should be visible in the playground
- You can toggle on and off an event listener per event
- Or you can turn on all events and have the listener available on page load
- Event names with there payloads are printed in the console

Widget Events tools

[LF-8931]: https://lifi.atlassian.net/browse/LF-8931?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/widget/src/stores/form/FormStore.tsx
```diff
@@ -130,6 +146,14 @@ export const FormStoreProvider: React.FC<FormStoreProviderProps> = ({
...
       ),
     );
+
+    // emit sendToWalletAddressChange event if toAddress is set in the config
+    if (toAddress) {
+      // debounce used to push this event emit to the back of the job queue
+      // otherwise without it the useEffect that is set up for listening to widget events
+      // can potentially be set up after the event has been emitted
+      debouncedToAddressWidgetEvent(toAddress.address, widgetEvents);
+    }
```
chybisov, line 156: Do you think we need to emit an event during initialization? I'm not a fan of debouncing logic here.

packages/widget/src/stores/form/useFieldActions.ts
```diff
@@ -1,8 +1,48 @@
...
 
+interface FieldValueToEmittedEvents {
+  [key: string]: (
+    value: GenericFormValue,
+    emitter: typeof widgetEvents,
+  ) => void;
+}
+
+const fieldValueToEmittedEvents: FieldValueToEmittedEvents = {
+  toAddress: (address, emitter) =>
```
chybisov, line 23: Let's add all fields here to avoid adding them in the future anyway.

packages/widget/src/types/events.ts
```diff
@@ -19,6 +19,7 @@ export enum WidgetEvent {
   WalletConnected = 'walletConnected',
   WidgetExpanded = 'widgetExpanded',
   PageEntered = 'pageEntered',
+  SendToWalletAddressChanged = 'sendToWalletAddressChanged',
```
chybisov, line 22: If we add all fields, we should come up with one event that has a field name and value. Wdyt?

Maybe something like 

- FormFieldChanged
- FormValueChanged

chybisov (changes requested):

packages/widget/src/stores/form/useFieldActions.ts
```diff
@@ -16,5 +56,43 @@ export const useFieldActions = () => {
...
+      options?: SetOptions,
+    ) => {
+      emitEventForFieldValueChange(
+        fieldName,
+        value,
+        actions.getFieldValues(fieldName)[0],
+        emitter,
+      );
+
+      actions.setFieldValue(fieldName, value, options);
```
chybisov, line 72: Should we set value first and, after that fire an event?

packages/widget/src/stores/form/useFieldActions.ts
```diff
@@ -16,5 +56,43 @@ export const useFieldActions = () => {
...
+      (Object.keys(formValues) as FormFieldNames[]).forEach((fieldName) => {
+        emitEventForFieldValueChange(
+          fieldName,
+          formValues[fieldName],
+          actions.getFieldValues(fieldName)[0],
+          emitter,
+        );
+      });
+
+      actions.setUserAndDefaultValues(formValues);
```
chybisov, line 88: Should we set value first and, after that fire an event?

packages/widget/src/stores/form/FormStore.tsx
DNR500: I'm happy to remove it if you don't think its needed

packages/widget/src/stores/form/useFieldActions.ts
DNR500: The actions.getFieldValues will not have the old value to do the comparison to check for change if you call actions.setFieldValue first. I tried it the other way around initially

packages/widget/src/stores/form/useFieldActions.ts
DNR500: The actions.getFieldValues will not have the old value to do the comparison to check for change if you call actions.setFieldValue first. I tried it the other way around initially

packages/widget/src/types/events.ts
DNR500: Makes sense, when doing this I did wonder it we would end you wanting to do it for all values

packages/widget/src/stores/form/useFieldActions.ts
DNR500: Lets go with one event for them all. It's simpler that way

packages/widget/src/stores/form/useFieldActions.ts
chybisov: I guess we can save it to a variable like 

`const oldValue = actions.getFieldValues(fieldName)[0]`

and then execute after.

packages/widget/src/stores/form/FormStore.tsx
chybisov: yes, please 🙂

packages/widget/src/stores/form/useFieldActions.ts
DNR500: That would actually become a bit more complicated in the example above for the `setUserAndDefaultValuesWithEmittedEvents` - you would have to store the oldValues in another object map first and then once `setUserAndDefaultValues` is executed you would have to the iterate over the old values map.

I don't see what the benefit will be of changing the order of execution there and I think it will make the code a little clean - but I do it and can decide if you are happier with than

packages/widget/src/stores/form/FormStore.tsx
DNR500: Have remove this

packages/widget/src/types/events.ts
DNR500: have gone with this approach using FormFieldChanged, I'm gonna look at better type inference as well. Also I think I used value rather than newValue so will change that

packages/widget/src/stores/form/useFieldActions.ts
DNR500: Have put something in place for this

packages/widget/src/stores/form/useFieldActions.ts
DNR500: Just gonna add in better type inference too

packages/widget/src/types/events.ts
DNR500: I've improved the types now as well

chybisov (approved):

PR #312, opened 2024-10-15
Settings event emit

DNR500:
wip


packages/widget/src/pages/SettingsPage/SlippageSettings/SlippageSettings.tsx
```diff
@@ -24,36 +22,48 @@ export const SlippageSettings: React.FC = () => {
...
+
+      setInputValue(formatSlippage(value, defaultValue.current, true));
+
+      debouncedSetValue(
+        'slippage',
+        formatSlippage(value || defaultSlippage, defaultValue.current, true),
+      );
+    },
+    [debouncedSetValue],
+  );
```
DNR500, line 52: Had to decouple the inputs value from setting the slippage value in the settings store. the debounce means this slippage value happens once and not on each keystroke

chybisov (approved):

PR #351, opened 2025-02-11
fix: transaction history page virtualization

tomiiide:
closes [LF-10831](https://lifi.atlassian.net/browse/LF-10831?atlOrigin=eyJpIjoiYjc4N2JhMjg4ZDJiNDVhMzhmNWY3MTg5ZWI3OGQ2YTkiLCJwIjoiaiJ9)

This PR fixes the issues with virtualization of the transaction history page.

### Technical Details
The problem was caused by the `flex: 1` property of the `Container` element used to render parent of the virtualized list.
This broke `react-virtual`  as the `Container` grew to the size of the child list, the entire user transaction history was being rendered instead of the number of elements that could fit into the fixed container height.

I switched the parent element from a  `Container` out for a `Box` component to fix it.

Additionally, I applied the `useListHeight` hook on the list so the height grew dynamically for when we switch from compact to drawer mode.

https://github.com/user-attachments/assets/260e7767-c77c-42f2-96b7-e74917876063

### DIFF
--- a/packages/widget/src/pages/TransactionHistoryPage/TransactionHistoryPage.tsx
+++ b/packages/widget/src/pages/TransactionHistoryPage/TransactionHistoryPage.tsx
@@ -1,6 +1,7 @@
 import type { FullStatusData } from '@lifi/sdk'
-import { List } from '@mui/material'
+import { Box, List } from '@mui/material'
 import { useVirtualizer } from '@tanstack/react-virtual'
+import React from 'react'
 import { useRef } from 'react'
 import { useTranslation } from 'react-i18next'
 import { PageContainer } from '../../components/PageContainer.js'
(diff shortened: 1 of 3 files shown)


packages/widget/src/pages/TransactionHistoryPage/TransactionHistoryPage.tsx
```diff
@@ -34,37 +41,48 @@ export const TransactionHistoryPage: React.FC = () => {
...
-          {Array.from({ length: 3 }).map((_, index) => (
-            <TransactionHistoryItemSkeleton key={index} />
-          ))}
-        </List>
-      ) : (
-        <List
+    <PageContainer disableGutters>
+      <Box style={{ minHeight: minTransactionListHeight }}>
+        <Box
+          ref={parentRef}
```
chybisov, line 47: Is it possible to reduce nesting here to two elements?

chybisov:
Other than that looks really smooth 🚀

PR #353, opened 2025-02-19
feat: widget reown example

tomiiide:
This PR creates an example of how to use the widget with AppKit from Reown app library.


examples/reown/src/providers/SyncedWalletProvider.tsx
```diff
@@ -0,0 +1,95 @@
...
+        const { chainNamespace } = caipNetwork
+        if (chainNamespace === 'solana') {
+          const connectors = appKit.getConnectors(chainNamespace)
+          // there's no way to get the active connector from appKit yet.
+          emitter.emit('connect', connectors[0].name)
+        } else {
+          emitter.emit('disconnect')
+        }
+      }
+    })
```
chybisov, line 64: Should we put this subscription inside useEffect?

examples/reown/src/config/appkit.ts
```diff
@@ -0,0 +1,10 @@
+export const projectId = import.meta.env.VITE_REOWN_APP_ID
+
+export const metadata = {
+  name: 'LiFi Widget',
+  description: 'Li.Fi Widget AppKit Example',
+  url: 'https://li.fi',
+  icons: [
+    'https://cdn.brandfetch.io/li.finance/w/512/h/195/logo?c=1id_Z_xJnoSm1uCNb-X',
```
chybisov, line 8: Can we serve the icon from the example repository?

examples/reown/src/providers/SyncedWalletProvider.tsx
```diff
@@ -0,0 +1,95 @@
...
+}: { children: React.ReactNode }) {
+  // fetch available chains before rendering the WalletProvider
+  const { chains, isLoading } = useAvailableChains()
+
+  if (!chains && isLoading) {
+    return null
+  }
+
+  return (
+    chains && (
```
chybisov, line 89: I see we already checked for chains in the early return above, so this might be unnecessary and we can use early return only, maybe extend it a bit to check for chains array length

examples/reown/src/utils/appkit.ts
```diff
@@ -0,0 +1,43 @@
+import { ChainType, type ExtendedChain } from '@lifi/widget'
+import type { AppKitNetwork, ChainNamespace } from '@reown/appkit-common'
+import { defineChain } from '@reown/appkit/networks'
+
+export type AppkitsupportedChainTypes = Exclude<ChainType, ChainType.MVM>
```
chybisov, line 5: small typo, `AppKitSupportedChainTypes`?

chybisov: Also, it might be better to only include supported AppKit chains to avoid updating this type when new ecosystems arrive.

examples/reown/README.md
```diff
@@ -0,0 +1,10 @@
+# Example Project for using LiFi widget with the AppKit from Reown
+
+This examples fetches chains information from Li.Fi using the `useAvailableChains` hook, and syncs wagmi config with the `useSyncWagmiConfig` hook.
```
chybisov, line 3: Please use `LI.FI` as the brand name casing 😁

examples/reown/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+
+# Installation
+1. Clone this repo
+2. install depedencies `pnpm install`
+3. rename `.env.example` to `.env`, and update env variables with your Reown APP ID
+4. run the app, `pnpm dev`
```
chybisov, line 10: Let's be more official in these docs (no typos, full terms, etc.):

# Installation
1. Clone example repository
2. Install dependencies `pnpm install`
3. Rename `.env.example` to `.env` and update environment variables with your Reown APP ID
4. Run the app `pnpm dev`

chybisov (changes requested):

examples/reown/src/providers/SyncedWalletProvider.tsx
tomiiide: Yeah, `subscribeCaipNetworkChange` doesn't return an unsubscribe cleanup function, but I found `subscribeNetwork` that serve the same function and returns an unsubscribe.

examples/reown/src/providers/SyncedWalletProvider.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { SolanaProvider, emitter } from './SolanaProvider'
+
+export function WalletProvider({
+  children,
+  chains,
+}: { children: React.ReactNode; chains: ExtendedChain[] }) {
+  const wagmi = useRef<WagmiAdapter | undefined>(undefined)
+  const modal = useRef<AppKit | undefined>(undefined)
+
+  const networks: [AppKitNetwork, ...AppKitNetwork[]] = [solana, bitcoin]
```
chybisov, line 23: Should we move this under the below `if` statement if it's not used outside of it anyway?

PR #358, opened 2025-03-03
fix(timers): display hours and days section of timers, and update copy

tomiiide:
Closes [LF-12067](https://lifi.atlassian.net/browse/LF-12607)
Closes [LF-12616](https://lifi.atlassian.net/browse/LF-12616)

This PR:
- fixes the timer bug where only display the minutes and seconds parts of the timer, now we display days, hours, minutes and seconds
- Removes the timer tool tip text to reflect the change from a count up timer to countdown timer

Screenshot 2025-03-03 at 11 48 17 AM

[LF-12067]: https://lifi.atlassian.net/browse/LF-12067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[LF-12616]: https://lifi.atlassian.net/browse/LF-12616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/widget/src/utils/timer.ts
+++ b/packages/widget/src/utils/timer.ts
@@ -1,0 +1,17 @@
+export const formatTimer = ({
+  days = 0,
+  hours = 0,
+  minutes = 0,
+  seconds = 0,
+}: {
+  days?: number
+  hours?: number
+  minutes?: number
+  seconds?: number
+}): string => {
+  const daysText = days > 0 ? (days < 10 ? `0${days}:` : `${days}:`) : ''
+  const hoursText = hours > 0 ? (hours < 10 ? `0${hours}:` : `${hours}:`) : ''
+  const minutesText = minutes < 10 ? `0${minutes}:` : `${minutes}:`
+  const secondsText = seconds < 10 ? `0${seconds} ` : `${seconds}`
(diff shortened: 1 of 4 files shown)


packages/widget/src/utils/timer.ts
tomiiide, line 1: @chybisov 

The new [Intl.DurationFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat) API would be appropriate here but it's not yet fully supported on all browsers yes.

Do we use a polyfill or keep it simple and wait until there is full support?

LMK what you think.

chybisov: Let's use the new API and also add a check if it is supported in the browser and if it is not supported we just don't show the timer. It's mostly adopted, but to avoid polyfills let's add a check.

https://caniuse.com/mdn-javascript_builtins_intl_durationformat_format

chybisov (approved):

PR #369, opened 2025-04-08
revert(timer): use a countdown instead of a count up

tomiiide:
https://lifi.atlassian.net/browse/LF-12981

Just going back to the old way.

https://github.com/user-attachments/assets/24f23bc9-1407-465c-a380-2c00ce1b467a


chybisov (changes requested):
Hey, in the video at 0:19, when we return to the main page, the timer is missing in the active routes (it should be visible). Then, when the transaction page is reopened, the timer restarts from the beginning, but it should continue counting down from where it left off.

chybisov:
@tomiiide Hey, I found one weird inconsistency - when we start the execution and the first time we get back to the main page, the timer slashes 5 seconds for some reason, but all consecutive page switches handle the time correctly. Please see video on 0:10-0:12.

https://github.com/user-attachments/assets/6c7e7bb7-dd01-4719-94c9-1e4d5b36d214

tomiiide:
https://github.com/lifinance/widget/pull/369#issuecomment-2796528030

@chybisov , I've fixed the issues in this comment, thanks for flagging it.

PR #390, opened 2025-05-06
fix(avatar-icon): add loading skeleton for loading image

tomiiide:
https://lifi.atlassian.net/browse/LF-13497

The reason for the empty region is that the image is being loaded from it's url before it's rendered, so I added a background image to the `<Avatar/>` to feel like a skeleton before the image is rendered.

https://github.com/user-attachments/assets/659e34af-5a1c-48f5-b9f7-5ab334e63234


packages/widget/src/components/TokenList/TokenListItem.tsx
```diff
@@ -101,7 +102,11 @@ export const TokenListItemButton: React.FC<TokenListItemButtonProps> = ({
       dense
     >
       <ListItemAvatar>
-        <Avatar src={token.logoURI} alt={token.symbol}>
+        <Avatar
+          src={token.logoURI}
+          alt={token.symbol}
+          sx={{ bgcolor: grey[400] }}
```
chybisov, line 108: We can't apply a background to all icons here because icons with transparent backgrounds (like wrapped SOL) will end up with an unintended grey background which looks off, see the screenshot for reference.

image

chybisov (changes requested):

packages/widget/src/components/TokenList/TokenListItem.tsx
```diff
@@ -101,7 +103,12 @@ export const TokenListItemButton: React.FC<TokenListItemButtonProps> = ({
       dense
     >
       <ListItemAvatar>
-        <Avatar src={token.logoURI} alt={token.symbol}>
+        <Avatar
+          src={token.logoURI}
+          alt={token.symbol}
+          sx={!imageLoaded ? { bgcolor: grey[400] } : null}
+          onLoad={() => setImageLoaded(true)}
```
chybisov, line 110: How does this affect performance? With this change, we will have 2x renders for every component in the list.

packages/widget/src/components/TokenList/TokenListItem.tsx
```diff
@@ -101,7 +103,12 @@ export const TokenListItemButton: React.FC<TokenListItemButtonProps> = ({
       dense
     >
       <ListItemAvatar>
-        <Avatar src={token.logoURI} alt={token.symbol}>
+        <Avatar
+          src={token.logoURI}
+          alt={token.symbol}
+          sx={!imageLoaded ? { bgcolor: grey[400] } : null}
```
chybisov, line 109: For consistency, it would be nice to use theme.vars with grey.300 here so it can be customized. We use grey.300 in other places for avatar placeholders, and we don't really use grey.400 for anything.

chybisov (changes requested):

packages/widget/src/components/TokenList/TokenListItem.tsx
chybisov: Wouldn't it re-render the whole <TokenListItemButton> component when setImageLoaded is triggered? Maybe we can add a `ListItemAvatarBase` component below and move `<Avatar>` together with `imageLoaded`/`setImageLoaded` state there, so we are sure we are updating only the small `<Avatar>` component, and not the whole `<TokenListItemButton>`.

tomiiide: Yeah, the entire component would be re-rendered, but I was testing with repaint flashing enabled and only the Avatar section was being repainted.

chybisov (approved):

PR #398, opened 2025-05-23
fix: height of virtualized list in drawer

effie-ms:
https://lifi.atlassian.net/browse/LF-13825

_`maxHeight` is determined differently for drawer subvariant, and that condition for drawer was missing for `height` in `.full-page-container` (which corresponds to general `maxHeight`)_ 

_Observe the list in the drawer takes the whole available height._  
https://github.com/user-attachments/assets/a2c6c114-c462-4072-9c3a-c8c7da18c324

### DIFF
--- a/packages/widget/src/App.tsx
+++ b/packages/widget/src/App.tsx
@@ -13,8 +13,8 @@ export const App = forwardRef<WidgetDrawer, WidgetProps>((props, ref) => {
       config.theme = {
         ...config.theme,
         container: {
-          height: '100%',
           ...config.theme?.container,
+          height: '100%',
         },
       }
     }
--- a/packages/widget/src/App.tsx
+++ b/packages/widget/src/App.tsx
@@ -13,8 +13,8 @@ export const App = forwardRef<WidgetDrawer, WidgetProps>((props, ref) => {
       config.theme = {
         ...config.theme,
         container: {
-          ...config.theme?.container,
           height: '100%',
+          ...config.theme?.container,
         },
       }
     }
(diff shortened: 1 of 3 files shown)


packages/widget/src/components/AppContainer.tsx
```diff
@@ -83,29 +83,30 @@ interface CssBaselineContainerProps {
...
-})<CssBaselineContainerProps>(({ theme, variant, paddingTopAdjustment }) => ({
-  display: 'flex',
-  flex: 1,
-  flexDirection: 'column',
-  overflowX: 'clip',
-  margin: 0,
-  width: '100%',
-  maxHeight:
+})<CssBaselineContainerProps>(({ theme, variant, paddingTopAdjustment }) => {
+  const fullPageHeight =
```
tomiiide, line 87: Nice extraction of logic to this const

tomiiide (approved):
Looks good!

chybisov (changes requested):
I'm not sure we can unify that logic as it break the behaviour we have for some lists. Please see the video, for the chains list we should have sticky header elements in the same way we have them on the playground or for transaction history.

https://github.com/user-attachments/assets/0c52f965-0bdc-4167-ab18-a2b619ffe3ae

packages/widget/src/App.tsx
```diff
@@ -13,8 +13,8 @@ export const App = forwardRef<WidgetDrawer, WidgetProps>((props, ref) => {
       config.theme = {
         ...config.theme,
         container: {
-          height: '100%',
           ...config.theme?.container,
+          height: '100%',
```
chybisov, line 17: Ah, nice catch! Looks like we override this `height: '100%'` with `height: undefined` that come from the playground.

image

Can we please fix this on the playground side? We need to have the ability for users to override `height: '100%'` in their configs, so this code block should stay the same as before and we need to find a place on the playground where we can adjust the config when the drawer variant is selected.

effie-ms: yeah, set it in `VariantControl` for playground, reverted the order for `App`, seems sufficient

chybisov (approved):
🚀

PR #401, opened 2025-05-26
Reuse react-query context for widget if available in react tree

effie-ms:
https://lifi.atlassian.net/browse/LF-13588

- [] If this PR modifies the Widget API, I have updated the documentation (or submitted a change request on GitBook).

### DIFF
--- a/packages/widget/src/AppProvider.tsx
+++ b/packages/widget/src/AppProvider.tsx
@@ -1,13 +1,13 @@
-import { QueryClientProvider } from '@tanstack/react-query'
+import { QueryClientContext, QueryClientProvider } from '@tanstack/react-query'
 import type { PropsWithChildren } from 'react'
-import { Fragment } from 'react'
+import { Fragment, useContext } from 'react'
 import {
   MemoryRouter,
   type MemoryRouterProps,
   useInRouterContext,
 } from 'react-router-dom'
 import { PageEntered } from './components/PageEntered.js'
-import { queryClient } from './config/queryClient.js'
+import { queryClient as newQueryClient } from './config/queryClient.js'
 import { I18nProvider } from './providers/I18nProvider/I18nProvider.js'
 import { ThemeProvider } from './providers/ThemeProvider/ThemeProvider.js'
 import { WalletProvider } from './providers/WalletProvider/WalletProvider.js'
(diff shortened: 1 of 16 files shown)


packages/widget/src/AppProvider.tsx
```diff
@@ -24,8 +24,9 @@ export const AppProvider: React.FC<PropsWithChildren<WidgetConfigProps>> = ({
   config,
   formRef,
 }) => {
+  const existingQueryClient = useContext(QueryClientContext)
   return (
-    <QueryClientProvider client={queryClient}>
+    <QueryClientProvider client={existingQueryClient || newQueryClient}>
```
chybisov, line 29: We should not render <QueryClientProvider> if this provider exists above in the React tree. It would be nice to encapsulate its rendering logic in a separate provider component.

packages/widget/src/hooks/useAvailableChains.ts
```diff
@@ -22,7 +23,7 @@ export const useAvailableChains = (chainTypes?: ChainType[]) => {
   // const { providers } = useHasExternalWalletProvider();
   const { data, isLoading } = useQuery({
     queryKey: [
-      'chains',
+      `${lifiWidgetQueryPrefix}-chains`,
```
chybisov, line 26: We have a customizable `keyPrefix` in the widget config, which should be taken into account here and other similar places I guess.

chybisov (changes requested):

packages/widget/src/AppProvider.tsx
effie-ms: Makes sense, thanks! 
Updated

packages/widget/src/AppProvider.tsx
```diff
@@ -37,7 +37,20 @@ export const AppProvider: React.FC<PropsWithChildren<WidgetConfigProps>> = ({
...
+    </ValidatedQueryClientProvider>
+  )
+}
+
+const ValidatedQueryClientProvider = ({
+  children,
+}: {
+  children: React.ReactNode
+}) => {
+  const hasQueryClient = !!useContext(QueryClientContext)
```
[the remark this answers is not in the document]
effie-ms: Updated

packages/widget/src/AppProvider.tsx
```diff
@@ -37,7 +37,20 @@ export const AppProvider: React.FC<PropsWithChildren<WidgetConfigProps>> = ({
...
+  return hasQueryClient ? (
+    children
+  ) : (
+    <QueryClientProvider client={queryClient}>{children}</QueryClientProvider>
   )
```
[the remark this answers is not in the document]
effie-ms: Yeah, let's do it, moved

chybisov (approved):

PR #404, opened 2025-05-27
feat: customization for heights

effie-ms:
https://lifi.atlassian.net/browse/LF-13552

_We introduce `height: fit-content` for widget container, to show widget content in full height (of the content). Since we still have long pages, they are restricted by `maxHeight` (separately from shorter pages). See added "Fit Content" option on playground._  

https://github.com/user-attachments/assets/46ca0f42-c90a-4b4b-b4e9-aebc8de55080


effie-ms:
> Actually found one issue, sent in DM
Resolved, I will ping QA

packages/widget/src/components/AppContainer.tsx
```diff
@@ -97,13 +107,19 @@ const CssBaselineContainer = styled(ScopedCssBaseline, {
...
     '&:has(.full-page-container)': {
-      height: fullContainerHeight,
+      height:
+        theme.container?.maxHeight ||
+        theme.container?.height ||
+        defaultMaxHeight,
+    },
+    '&:has(.long-list)': {
+      maxHeight: theme.container?.maxHeight || defaultMaxHeight,
     },
```
tomiiide, line 123: Is there another way to do this without having to add a class to every "long-list" in the codebase?
My problem with this is if the widget grows, it would be a pain to manage.

packages/widget/src/pages/RoutesPage/RoutesPage.tsx
```diff
@@ -69,7 +69,7 @@ export const RoutesPage: React.FC<BoxProps> = () => {
   const allowInteraction = account.isConnected && !toAddressUnsatisfied
 
   return (
-    <Stack direction="column" spacing={2} flex={1}>
+    <Stack className="long-list" direction="column" spacing={2} flex={1}>
```
tomiiide, line 72: If we have to do this, can we have a `<LongList>` component or HOC that does handles the className?

packages/widget/src/components/AppContainer.tsx
chybisov: This is actually the best way to do this, because it uses native CSS features, instead of some JS ad-hoc solutions.

chybisov (approved):

PR #414, opened 2025-06-02
feat: add optional filter for wallets in modal

effie-ms:
https://lifi.atlassian.net/browse/LF-12735

Adding optional arguments `{ chain, chainType }` as described in the task. 
1) Not enabling support for `chain: number`, as retrieving chain by id needs widget configuration (see useAvailableChains()), which creates a circular dependency (if not copy-pasting the code)
2) Filtering wallets only by type, as this is the only filter existing for wallets  (see types `CombinedWalletConnector` and `CombinedWallet`)

 https://github.com/user-attachments/assets/00f2cab6-0530-41a2-a191-94698c36434d


packages/wallet-management/src/providers/WalletMenuProvider/WalletMenuContext.ts
```diff
@@ -1,10 +1,13 @@
...
+import type {
+  WalletMenuOpenArgs,
+  WalletMenuContext as _WalletMenuContext,
+} from './types.js'
 
 export const WalletMenuContext = createContext<_WalletMenuContext>({
   isWalletMenuOpen: () => {},
   toggleWalletMenu: () => {},
-  openWalletMenu: () => {},
+  openWalletMenu: (_?: WalletMenuOpenArgs) => {},
```
chybisov, line 10: Can we please name it `args` instead of `_`

packages/wallet-management/src/i18n/en.json
```diff
@@ -1,6 +1,7 @@
 {
   "title": {
     "connectWallet": "Connect wallet",
+    "connectWalletWithChain": "Connect {{chainLabel}} wallet",
```
chybisov: Let' use `chainName` here, we call it like this in other places

packages/wallet-management/src/providers/WalletMenuProvider/WalletMenuContext.ts
effie-ms: this rule gets triggered then: https://biomejs.dev/linter/rules/no-unused-variables/

chybisov: Ah, okay 👌 I haven't noticed, it's not a type

packages/wallet-management/src/i18n/en.json
effie-ms: It is chain name or type (EVM, etc) there currently, should we still call them "chainName"?

chybisov: Ah, I see, let's leave `chainLabel` then 👍

tomiiide (approved):
Looks good!

effie-ms:
@chybisov as agreed - adding only ecosystem filter (to preselect ecosystem if chainType filter exists) from QA suggestions:
https://github.com/lifinance/widget/pull/414/commits/a3a8de890a4cb1cdd391c29b6422287c93a92551

chybisov (approved):

PR #418, opened 2025-06-04
feat: restyled connect wallet view

effie-ms:
https://lifi.atlassian.net/browse/LF-13931

Wallet connect modal is styled according to the new design.
Deviations from the design:
1. Using only 2 colors for tags, colors for “Installed” tag are currently not suitable for LiFi’s theme (“Default”) contrast-wise - needs pallette sync first.
2. No “Recent” tag (as per task spec, logic does not exist yet)
3. No “All wallets” (as per task spec)

Additional design requests:
- [x] Add "Connected" tag 
- [x] Darker (6%) hover for wallet cards

image
image
image


packages/wallet-management/src/components/WalletInfoDisplay.tsx
```diff
@@ -0,0 +1,59 @@
...
+        px: 1,
+      }}
+    >
+      <Avatar
+        src={selectedWallet?.icon}
+        alt={selectedWallet?.name}
+        sx={{ width: 96, height: 96 }}
+      >
+        {selectedWallet?.name[0]}
+      </Avatar>
```
tomiiide: Should we have a default value for the Avatar in case `selectedWallet` is `null`?

packages/wallet-management/src/components/WalletMenuContent.tsx
```diff
@@ -99,12 +101,43 @@ export const WalletMenuContent: React.FC<WalletMenuContentProps> = ({
...
+                  WalletTagType.Connected
+              )
+                ? WalletTagType.Connected
+                : WalletTagType.Multichain
+            } else if (wallet.connectors.length === 1) {
+              walletTagType = getTagType(
+                wallet.connectors[0].connector,
+                wallet.connectors[0].chainType
+              )
+            }
```
tomiiide: Would it better to move this logic to a util function too?

tomiiide (changes requested):
@effie-ms UI update looks good, I just need clarity on some code.

packages/wallet-management/src/components/WalletInfoDisplay.tsx
effie-ms: Theoretically, it is not supposed to happen on those views, but yeah, adding showing the avatar conditionally on existence, just in case

packages/wallet-management/src/components/WalletMenuContent.tsx
effie-ms: Yeah, good idea, moved to getWalletTagType()

PR #420, opened 2025-06-05
feat: new chain sidebar

effie-ms:
https://lifi.atlassian.net/browse/LF-13596

### DIFF
--- a/packages/widget/src/components/Expansion/Expansion.tsx
+++ b/packages/widget/src/components/Expansion/Expansion.tsx
@@ -1,0 +1,40 @@
+import { Collapse } from '@mui/material'
+import type { PropsWithChildren } from 'react'
+import type { RouteObject } from 'react-router-dom'
+import { useRoutes as useDOMRoutes } from 'react-router-dom'
+import { CollapseContainer, RouteTopLevelGrow } from './Expansion.style'
+
+interface ExpansionProps {
+  allowedPaths: string[]
+}
+
+const timeout = { enter: 225, exit: 225, appear: 0 }
+
+export const Expansion = ({
+  allowedPaths,
+  children,
+}: PropsWithChildren<ExpansionProps>) => {
+  const routes: RouteObject[] = [
+    ...allowedPaths.map((path) => ({ path: path, element: true })),
+    {
+      path: '*',
+      element: null,
+    },
+  ]
+  const element = useDOMRoutes(routes)
+  const match = Boolean((element?.props as PropsWithChildren)?.children)
+  return (
+    <CollapseContainer>
+      <Collapse timeout={timeout} in={match} orientation="horizontal">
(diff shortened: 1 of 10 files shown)


packages/widget/src/components/Expansion/Expansion.tsx
tomiiide, line 1: Good abstraction

packages/widget/src/components/Search/SearchInput.tsx
```diff
@@ -21,17 +22,27 @@ export const SearchInput = ({
...
+          }
           endAdornment={
-            <InputAdornment position="end">
-              <Search />
-            </InputAdornment>
+            inExpansion ? null : (
+              <InputAdornment position="end">
+                <Search />
+              </InputAdornment>
+            )
```
tomiiide: Since the `SearchInput` would be used in many different contexts, I think it's better to accept Input props directly from its parent, so it can be customized as needed.

So instead of `size={inExpansion ? 'small' : 'medium"`, we can just pass in the `size` directly , same for other props.

What do you think?

effie-ms: Yeah, I see your point. `inExpansion` is used for adornments (search icon) as well, maybe then it makes sense to pass this as a prop as well. 
Making 2 props there for `SearchInput`: `size` ( small | medium) and `iconPosition` (start | end), this should do the job without `inExpansion`. Semantically it does not make sense to have `inExpansion` for the input indeed.

packages/widget/src/components/Search/SearchInput.tsx
```diff
@@ -54,7 +67,7 @@ export const StickySearchInput = (props: SearchInputProps) => {
 
   return (
     <StickySearchInputContainer headerHeight={headerHeight}>
-      <SearchInput {...props} autoFocus />
+      <SearchInput {...props} autoFocus size="medium" iconPosition="end" />
```
tomiiide: No need for this change as the defaults handle this case.

packages/widget/src/pages/SelectTokenPage/SearchTokenInput.tsx
```diff
@@ -25,6 +25,9 @@ export const SearchTokenInput = () => {
       onChange={(e) => onChange((e.target as HTMLInputElement).value)}
       onBlur={onBlur}
       value={value as string | undefined}
+      autoFocus
+      size="medium"
+      iconPosition="end"
```
tomiiide: Same here, no need for the defaults values for `size` and `iconPosition`

vinzenzLIFI (approved):
lgtm. I like the gap between the main widget and the chain list. works on both light and dark mode. The different layouts also look correct with the new side bar. 
As discussed this diverts from the design, but it's already known and discussed with design.

PR #447, opened 2025-07-01
fix(widget): remove always on reconnection for bitcoin wallets

tomiiide:
There is no need for the `useReconnect` hook as the `reconnectOnMount` param on provider component has been configured to `false`


chybisov:
@tomiiide 

> There is no need for the useReconnect hook as the reconnectOnMount param on provider component has been configured to false

This is not like it seems. We have `reconnectOnMount` as `false` and an explicit `useReconnect` hook to avoid Next.js and other mounting issues with Wagmi. The same setup was migrated to Bigmi, but I'm not sure if that's needed, and maybe we can safely just enable `reconnectOnMount`. We can't remove them both, because in that case, when you connect the wallet and refresh the page, the connection won't restore automatically.

tomiiide:
@chybisov You're right, this is not the root cause of the problem. The problem is addressed in this PR.
https://github.com/lifinance/bigmi/pull/24/files

PR #451, opened 2025-07-04
feat: optimize lookups for tokens

effie-ms:
https://lifi.atlassian.net/browse/LF-14494

This PR optimizes complexity of filters by adding sets and performing O(1) search instead of O(n) filtering.
Needed to speed up filtering of long lists for smoother UX (especially once we get super long lists like "All tokens")

### DIFF
--- a/packages/widget/src/hooks/useTokenBalances.ts
+++ b/packages/widget/src/hooks/useTokenBalances.ts
@@ -36,6 +36,7 @@ export const useTokenBalances = (
       account.address,
       selectedChainId,
       tokens?.length,
+      formType,
     ],
     queryFn: async ({ queryKey: [, accountAddress] }) => {
       const tokensWithBalance: TokenAmount[] = await getTokenBalances(


packages/widget/src/types/widget.ts
```diff
@@ -209,19 +209,30 @@ export interface AllowDeny<T> {
...
+export type AllowDenySets = {
+  from?: AllowDenySet
+  to?: AllowDenySet
+} & AllowDenySet
+
+export type AllowDenyItems<T> = {
+  from?: AllowDeny<T>
+  to?: AllowDeny<T>
+} & AllowDeny<T>
+
```
chybisov: Ah, we can't change this to Sets right away without making it a breaking change. We should have a convert function instead. We can possibly consider breaking change for v4.

effie-ms: The changes are analogous to what we had before, new types do not make a breaking change. 
Resolving

packages/widget/src/hooks/useAvailableChains.ts
```diff
@@ -33,9 +33,15 @@ export const useAvailableChains = (chainTypes?: ChainType[]) => {
       chains?.to,
     ] as const,
     queryFn: async ({ queryKey: [, chainTypesConfig] }) => {
+      const chainsConfigSets = getConfigItemSets(
+        chainTypesConfig,
+        (chains) => new Set(chains.map(String))
```
chybisov: Why do we need `.map(String)` here?

packages/widget/src/hooks/useAvailableChains.ts
```diff
@@ -33,9 +33,15 @@ export const useAvailableChains = (chainTypes?: ChainType[]) => {
...
+      )
       const chainTypesRequest = supportedChainTypes
         // providers.length > 0 ? providers : supportedChainTypes
-        .filter((chainType) => isItemAllowed(chainType, chainTypesConfig))
+        .filter((chainType) =>
+          isItemAllowedForSets(chainType, chainsConfigSets, String)
```
chybisov: Maybe it's better to make `getKey` function optional and use `item` as a key if getKey wasn't provided?

effie-ms: Nice, will be shorter, I have updated 👍

packages/widget/src/hooks/useTokens.ts
```diff
@@ -48,17 +48,27 @@ export const useTokens = (selectedChainId?: number, formType?: FormType) => {
...
-    const filteredConfigTokens = filterConfigTokensByChain(
+    const filteredConfigTokens = getConfigItemSets(
       configTokens,
-      formType,
-      selectedChainId
+      (tokens: BaseToken[]) =>
+        new Set(
+          tokens
+            .filter((t) => t.chainId === selectedChainId)
+            .map((t) => `${t.address}-${t.chainId}`)
```
chybisov: If we have already filtered tokens by chainId, why do we need to include chainId in the key? The same applies to the logic below, where we also check the chainId before checking `isFormItemAllowed`

vinzenzLIFI (approved):
tested on the current state. no regressions were found ✅
For details on the test, please see the jira ticket.

chybisov (approved):

PR #465, opened 2025-07-17
fix(format): wrap hashes in messages for better display

tomiiide:
https://lifi.atlassian.net/browse/LF-8638

The solution I explored was checking messages for hashes, and wrapping with middle elipses, that way, the full error message is visible, but the hashes don't overflow.

Screenshot 2025-07-17 at 4 37 08 PM
Screenshot 2025-07-17 at 4 36 59 PM

### DIFF
--- a/packages/widget/src/utils/format.ts
+++ b/packages/widget/src/utils/format.ts
@@ -114,3 +114,22 @@ export function formatDuration(seconds: number, locale: string): string {
     unitDisplay: 'narrow',
   })
 }
+
+export function wrapHashes(text: string): string {
+  const shouldWrap = (text: string): boolean => {
+    const cleanText = text.replace(/[.,!?;:]$/, '')
+    const hashPattern = /^[A-Za-z0-9]{32,}$/
+    return hashPattern.test(cleanText)
+  }
+  return text
+    .split(' ')
+    .map((word) => {
+      if (shouldWrap(word)) {
+        return word.length > 12
+          ? `${word.slice(0, 8)}...${word.slice(-4)}`
+          : word
+      }
+      return word
+    })
+    .join(' ')
+}
--- a/packages/widget/src/components/Step/StepProcess.tsx
+++ b/packages/widget/src/components/Step/StepProcess.tsx
@@ -10,7 +10,7 @@ export const StepProcess: React.FC<{
   step: LiFiStep
   process: Process
 }> = ({ step, process }) => {
-  const { title, message } = useProcessMessage(step, process)
+  const { title, messageWrapped } = useProcessMessage(step, process)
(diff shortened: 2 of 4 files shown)


packages/widget/src/utils/format.ts
```diff
@@ -114,3 +114,22 @@ export function formatDuration(seconds: number, locale: string): string {
...
+    .map((word) => {
+      if (shouldWrap(word)) {
+        return word.length > 12
+          ? `${word.slice(0, 8)}...${word.slice(-4)}`
+          : word
+      }
+      return word
+    })
+    .join(' ')
+}
```
chybisov: Here we check for 32+ character length hashes and then slice them when they are 12+ character length. I believe we can optimize this, wdyt? Maybe just checking the length would be enough?

chybisov (changes requested):

packages/widget/src/utils/format.ts
effie-ms: Maybe rather some sort of break-word there? 
Also looks like something to change in the SDK if we compile the whole message there, not parsing post factum

chybisov: The SDK doesn't know about error messages it throws and formatting should happen on the UI layer.

tomiiide: nice catch @chybisov

tomiiide: @chybisov  Just to clarify, do you mean we should not check for hashes with the regex, and just shorten any word with a length > 32?

chybisov: @tomiiide yes, I think regex can be skipped. @effie-ms wdyt?

effie-ms: I don't think there are many 32+ characters-long words in general, so it should be sufficient indeed, good idea

tomiiide: Makes sense @chybisov @effie-ms. It's done.

packages/widget/src/utils/format.ts
```diff
@@ -114,3 +114,12 @@ export function formatDuration(seconds: number, locale: string): string {
     unitDisplay: 'narrow',
   })
 }
+
+export function wrapLongWords(text: string): string {
```
effie-ms, line 118: with `replace` should be shorter, unless we have anything against regex here
```
text.replace(/\S{32,}/g, (word) => {
    return `${word.slice(0, 8)}...${word.slice(-4)}`
})
```

tomiiide: Oh, I didn't know of that function, yeah that is actually much cleaner

effie-ms (approved):

PR #472, opened 2025-07-21
chore: test the widget on all package managers

tomiiide:
https://lifi.atlassian.net/browse/LF-8778

An example project was created for each runtime as a showcase of how to configure each runtime to work with the widget.

The examples were excluded from the `pnpm` workspace to avoid package management clashes.


tomiiide:
Closing as example for package managers are not necessary

PR #473, opened 2025-07-21
fix: improve tx link handling for relayed txs

chybisov:
Moved the fix from the SDK to the Widget for tx link handling.


packages/widget/src/components/Step/StepProcess.tsx
```diff
@@ -13,6 +13,18 @@ export const StepProcess: React.FC<{
   const { title, message } = useProcessMessage(step, process)
   const { getTransactionLink } = useExplorer()
 
+  const transactionLink = process.txHash
```
effie-ms, line 16: If you consider to shorten it a bit
```
const transactionLink =  process.txHash || process.txLink ?  getTransactionLink({
  txHash: process.txHash,
  txLink: process.txLink,
  chain: process.chainId,
}) : undefined;
```

packages/widget/src/hooks/useExplorer.ts
```diff
@@ -50,19 +51,32 @@ export const useExplorer = () => {
...
     txHash,
     txLink,
     chain,
-  }: TransactionLinkProps) => {
+  }: TransactionLinkProps): string | undefined => {
     if (!txHash && txLink) {
       return txLink
     }
 
+    if (!txHash) {
```
effie-ms, line 63: (to replace the block above)
```
if (!txHash) {
  return txLink ?? undefined;
}
```

packages/widget/src/components/Step/StepProcess.tsx
chybisov: we can't do this because of types 😁

packages/widget/src/hooks/useExplorer.ts
chybisov: done ✅

tomiiide (changes requested):
@chybisov  conflicts needs to be resolved tho.

PR #475, opened 2025-07-22
feat: market cap and volume for tokens

effie-ms:
**_NB: needs TokenAmount type to be updated in SDK_**

http://lifi.atlassian.net/browse/LF-14675

Adding market cap and volume 24h, FDV seems to be null for tokens currently. 

image

### DIFF
--- a/packages/widget/src/types/token.ts
+++ b/packages/widget/src/types/token.ts
@@ -3,6 +3,4 @@ import type { TokenAmount as SDKTokenAmount } from '@lifi/sdk'
 export interface TokenAmount extends SDKTokenAmount {
   featured?: boolean
   popular?: boolean
-  marketCapUSD?: number | null
-  volumeUSD24H?: number | null
 }
--- a/packages/widget/src/hooks/useTokenSearch.ts
+++ b/packages/widget/src/hooks/useTokenSearch.ts
@@ -2,12 +2,12 @@ import {
   type BaseToken,
   type ChainId,
   getToken,
+  type TokenExtended,
   type TokensResponse,
 } from '@lifi/sdk'
 import { useQuery, useQueryClient } from '@tanstack/react-query'
 import { useWidgetConfig } from '../providers/WidgetProvider/WidgetProvider.js'
 import type { FormType } from '../stores/form/types.js'
-import type { TokenAmount } from '../types/token.js'
 import { getConfigItemSets, isFormItemAllowed } from '../utils/item.js'
 import { getQueryKey } from '../utils/queries.js'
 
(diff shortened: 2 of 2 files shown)


packages/widget/src/utils/format.ts
```diff
@@ -1,5 +1,18 @@
...
+ * @param amount amount to format.
+ * @returns formatted amount.
+ */
+export function formatAmount(amount: number | null | undefined): string {
+  if (typeof amount !== 'number' || Number.isNaN(amount)) {
+    return '0'
+  }
+
+  return amount.toString()
+}
```
chybisov: Do we really need this util func? I guess we already have formatting for USD values, the ones that we use with the standard currency formatter

packages/widget/src/utils/currencyExtendedFormatter.ts
```diff
@@ -13,3 +13,15 @@ export const currencyExtendedFormatter = (
...
+  lng: string | undefined,
+  options: any
+) => {
+  return currencyExtendedFormatter(lng, {
+    ...options,
+    notation: 'compact',
+    compactDisplay: 'short',
+    maximumFractionDigits: 2,
+  })
+}
```
chybisov: I assume we couldn't reuse the existing one and pass parameters via translations? 😁

effie-ms: Yes, it wants a simple string, and does not parse correctly when they are more than 1 option,
We could overwrite "interpolation" to add parsing there, but I think something like this (another formatting function) is easier to maintain eventually

packages/widget/src/utils/format.ts
effie-ms: Removed, seems to work fine without it as well

packages/widget/src/utils/currencyExtendedFormatter.ts
effie-ms: Fake news actually, you can pass it via t() as params, replaced already, sorry for desinformation

packages/widget/src/types/token.ts
```diff
@@ -3,4 +3,6 @@ import type { TokenAmount as SDKTokenAmount } from '@lifi/sdk'
 export interface TokenAmount extends SDKTokenAmount {
   featured?: boolean
   popular?: boolean
+  marketCapUSD?: number | null
+  volumeUSD24H?: number | null
```
chybisov: The latest SDK version includes the TokenExtended type, which we can inherit from here

effie-ms: updated

PR #480, opened 2025-07-23
feat: minFromAmountUSD

effie-ms:
https://lifi.atlassian.net/browse/LF-14545 

image


packages/widget/src/components/Messages/MinFromAmountUSDMessage.tsx
```diff
@@ -0,0 +1,36 @@
...
+      title={
+        <Typography
+          variant="body2"
+          sx={{
+            color: 'text.primary',
+          }}
+        >
+          {t('warning.message.minFromAmountUSD', {
+            amount: minFromAmountUSD,
+            currency: 'USD',
```
chybisov: USD should be the default one there, so probably we can omit this.

packages/widget/src/components/Messages/useMessageQueue.ts
```diff
@@ -52,24 +61,32 @@ export const useMessageQueue = (route?: Route, allowInteraction?: boolean) => {
       })
     }
 
+    if (isLowerThanMinFromAmountUSD) {
+      queue.push({
+        id: 'MIN_FROM_AMOUNT_USD',
+        priority: 4,
+        props: { minFromAmountUSD },
+      })
+    }
```
chybisov: It should probably be priority 3, after the insufficient funds.

packages/widget/src/components/Messages/useMessageQueue.ts
```diff
@@ -25,6 +26,14 @@ export const useMessageQueue = (route?: Route, allowInteraction?: boolean) => {
     useGasSufficiency(route)
   const { missingChain, missingAccountAddress } =
     useRouteRequiredAccountConnection(route)
+  const { minFromAmountUSD } = useWidgetConfig()
+
+  const isLowerThanMinFromAmountUSD = useMemo(() => {
+    if (!minFromAmountUSD || !route?.fromAmountUSD) {
```
chybisov: I don't think it makes sense to wait for the route to declare that the user doesn't meet this requirement. Please see how we handle insufficient funds.

Also, maybe we should make this part of the insufficient funds hook or create a new one to use here.

chybisov (changes requested):

packages/widget/src/components/Messages/useMessageQueue.ts
effie-ms: updated

packages/widget/src/components/Messages/MinFromAmountUSDMessage.tsx
effie-ms: updated

packages/widget/src/components/Messages/useMessageQueue.ts
effie-ms: updated: calculating from the field and price directly, moved the logic to a hook

packages/widget/src/components/Messages/useMessageQueue.ts
```diff
@@ -25,6 +26,8 @@ export const useMessageQueue = (route?: Route, allowInteraction?: boolean) => {
     useGasSufficiency(route)
   const { missingChain, missingAccountAddress } =
     useRouteRequiredAccountConnection(route)
+  const { insufficientFromAmountUSD, minFromAmountUSD } =
+    useFromAmountSufficiency()
```
chybisov: Maybe we should use something like `belowMinFromAmountUSD` and `useFromAmountThreshold`. Insufficiency implies an absolute failure like not enough balance. Wdyt? 😅

packages/widget-playground/src/defaultWidgetConfig.ts
```diff
@@ -7,6 +7,7 @@ export const widgetBaseConfig: WidgetConfig = {
   // fromToken: '0x2791bca1f2de4661ed88a30c99a7a9449aa84174',
   // toToken: '0x7f5c764cbc14f9669b88837ca1490cca17c31607', // 0x0000000000000000000000000000000000000000
   // fromAmount: '20',
+  minFromAmountUSD: 100,
```
chybisov: let's remove it, so the playground does not have a minimum amount 🙂

packages/widget/src/components/Messages/useMessageQueue.ts
effie-ms: Renamed 😄

PR #482, opened 2025-07-24
feat(wallet-management): add support for base account

tomiiide:
https://lifi.atlassian.net/browse/LF-14696

It was implemented the standard way we implement these kinds of wallets.

Screenshot 2025-07-29 at 11 31 50

### DIFF
--- a/packages/widget/src/hooks/useGasSufficiency.ts
+++ b/packages/widget/src/hooks/useGasSufficiency.ts
@@ -57,7 +57,7 @@ export const useGasSufficiency = (route?: RouteExtended) => {
       EVMAccount.chainType
     )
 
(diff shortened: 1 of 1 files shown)


chybisov (commented):
Can you please check if we still need Coinbase? 👀

tomiiide:
> Can you please check if we still need Coinbase? 👀

Yeah we do for one reason, it is the only way to connect with the Coinbase app wallet on mobile.

Screenshot 2025-07-29 at 12 38 50


However, the Coinbase wallet automatically connects to an existing base account connection, and also give you an option to login in to one during the auth flow. (which makes things a bit confusing)

Also, we should rename the Coinbase wallet to the Base App, and update the logo.

chybisov (approved):

PR #488, opened 2025-07-28
fix: delay clearing token details not to have empty state on transition

effie-ms:
Currently, when token details drawer closes, the state gets cleared immediately, meaning that on closing transition there can be noticed an empty state. This PR removes clearing the state.

Before:
Screenshot 2025-07-28 at 11 26 16
https://github.com/user-attachments/assets/7655bf82-fdcc-4f90-990d-21e6ac4b488a
After:
https://github.com/user-attachments/assets/490c3638-74af-4976-ad5c-8e757e959ca2

### DIFF
--- a/packages/widget/src/components/TokenList/TokenDetailsSheet.tsx
+++ b/packages/widget/src/components/TokenList/TokenDetailsSheet.tsx
@@ -13,9 +13,6 @@ export const TokenDetailsSheet = forwardRef<
   TokenDetailsSheetProps
 >(({ chainId }, ref) => {
   const bottomSheetRef = useRef<BottomSheetBase>(null)
-  const timeoutRef = useRef<ReturnType<typeof setTimeout> | undefined>(
-    undefined
-  )
   const [tokenAddress, setTokenAddress] = useState<string | undefined>(
     undefined
   )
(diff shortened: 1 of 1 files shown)


packages/widget/src/components/TokenList/TokenDetailsSheet.tsx
```diff
@@ -29,8 +32,15 @@ export const TokenDetailsSheet = forwardRef<
...
-        setWithoutContractAddress(false)
+        if (timeoutRef.current) {
+          clearTimeout(timeoutRef.current)
+        }
+        // Delay clearing state until after bottom sheet animation completes
+        timeoutRef.current = setTimeout(() => {
+          setTokenAddress(undefined)
+          setWithoutContractAddress(false)
+          timeoutRef.current = undefined
+        }, 200)
```
chybisov: Maybe we can just avoid setting the state to undefined? Next time, when we gonna open these details, we first set a new state. So the old one should not hurt 🤔

effie-ms: Makes sense indeed, would not hurt
Updated

chybisov (approved):

PR #493, opened 2025-07-31
refactor: useShallow usage for zustand stores

effie-ms:
https://lifi.atlassian.net/browse/LF-14768

This PR refactors used patterns for zustand stores.
The goal is to prevent unnecessary rerenders using `useShallow`.
In particular, we migrate from manual `shallow` comparisons to using `useShallow` from `zustand/shallow` for all store hooks - to automatically applying shallow comparison to selectors.
All the widget stores now use `useShallow` internally and all the corresponding hook usages do not need passing `shallow`.

```typescript
// Before
export function useBookmarkStore<T>(
  selector: (state: BookmarkState) => T,
  equalityFn = shallow
): T {
  const useStore = useContext(BookmarkStoreContext)
  return useStore(selector, equalityFn)
}

// After
export function useBookmarkStore<T>(
  selector: (state: BookmarkState) => T
): T {
  const useStore = useContext(BookmarkStoreContext)
  return useStore(useShallow(selector))
}
```

```typescript
// Before: Manual shallow comparison
const [bookmarks, selectedBookmark] = useBookmarkStore(
  (store) => [store.bookmarks, store.selectedBookmark],
  shallow
)

// After: Automatic useShallow
const [bookmarks, selectedBookmark] = useBookmarkStore(
  (store) => [store.bookmarks, store.selectedBookmark]
)
```

Now all the stores follow the same optimization pattern.


examples/zustand-widget-config/src/store/createWidgetConfigStore.ts
```diff
@@ -47,5 +47,11 @@ export const useWidgetConfigStore = createWithEqualityFn<WidgetConfigState>(
...
   }),
-  shallow
+  Object.is
 )
+
+export const useWidgetConfigStore = <T>(
+  selector: (state: WidgetConfigState) => T
+): T => {
+  return createWidgetConfigStore(useShallow(selector))
+}
```
chybisov: Looks like in this case, we are going to create a store on every call to `useWidgetConfigStore`, which is kinda not optimal 🤔

examples/zustand-widget-config/src/store/createWidgetConfigStore.ts
```diff
@@ -47,5 +47,11 @@ export const useWidgetConfigStore = createWithEqualityFn<WidgetConfigState>(
       })
     },
   }),
-  shallow
+  Object.is
```
chybisov, line 50: Why did we change this to Object.is?

packages/widget-playground/src/hooks/useDevView.ts
```diff
@@ -1,12 +1,11 @@
-import { shallow } from 'zustand/shallow'
 import { useEditToolsStore } from '../store/editTools/EditToolsProvider'
 import { useEditToolsActions } from '../store/editTools/useEditToolsActions'
 import { setQueryStringParam } from '../utils/setQueryStringParam'
 
 const queryStringKey = 'devView'
 
 export const useDevView = () => {
-  const [isDevView] = useEditToolsStore((store) => [store.isDevView], shallow)
+  const [isDevView] = useEditToolsStore((store) => [store.isDevView])
```
chybisov: When taking one value from the store, there is no point in wrapping it in an array, simpler would be to have it like this:

```
const isDevView = useEditToolsStore((store) => store.isDevView)
```

packages/widget-playground/src/store/editTools/EditToolsProvider.tsx
```diff
@@ -48,5 +45,5 @@ export function useEditToolsStore<T>(
     )
   }
 
-  return useStore(selector, equalityFunction)
+  return useStore(useShallow(selector))
```
chybisov, line 48: Are you sure we need to use useShallow here because in this case it will be applied to every usage of the store, and in some cases we might not want this. Wdyt?

packages/widget-playground/src/store/editTools/useCodeToolValues.ts
```diff
@@ -1,11 +1,9 @@
...
 import { useEditToolsStore } from './EditToolsProvider'
 
 export const useCodeToolValues = () => {
-  const [codeControlTab] = useEditToolsStore(
-    (store) => [store.codeControl.openTab],
-    shallow
-  )
+  const [codeControlTab] = useEditToolsStore((store) => [
+    store.codeControl.openTab,
+  ])
```
chybisov: Same one property, we can skip the array

examples/zustand-widget-config/src/store/createWidgetConfigStore.ts
chybisov: Ah, we just renamed the store, a bit confusing now 😅

packages/widget-playground/src/hooks/useDevView.ts
effie-ms: Updated all the single values

packages/widget-playground/src/store/editTools/EditToolsProvider.tsx
effie-ms: We do have a few single values that would be fine with === instead of `useShallow`, but I think using the `useShallow for them is a negligible overhead, just choosing consistency in this particular case

packages/widget-playground/src/store/editTools/useCodeToolValues.ts
effie-ms: updated

examples/zustand-widget-config/src/store/createWidgetConfigStore.ts
effie-ms: this is the equality check we use for all the `createWithEqualityFn` in all the other stores

examples/zustand-widget-config/src/store/createWidgetConfigStore.ts
effie-ms: Yeah, they are just singletons, not recreated. I have renamed (e.g. `widgetConfigStore(useShallow(selector))`) those that I updated, though we have them called "create..." in all the other cases as well.

chybisov (approved):

PR #498, opened 2025-08-04
fix: resolve drawer/modal focus traps

effie-ms:
Problem: We have several accessibility issues with modal-like components. One consistently reproducible issue occurs with drawers displaying token details. When using interactive elements within a drawer and then closing it, the focused element remains focused at the moment aria-hidden="true" is applied to the drawer.
One considered solution was to set disableAutoFocus to true for MUI’s Drawer. However, focus is still trapped, and the focused drawer element visually glitches, jumping to the center.
Suggested solution: Apply the inert property to the drawer before closing it. This also requires separating the batched state update by wrapping the actual close event in a setTimeout, so that the state updates in different ticks.
Occasionally, clicking the info button itself causes the click to be trapped while the drawer appears on top. This issue can be resolved by blurring the button element before opening the drawer.  

## Visual showcase (to reproduce):  

https://github.com/user-attachments/assets/5100573c-4b6c-46df-a45f-1c865b4e63bc

https://github.com/user-attachments/assets/061096bb-e20f-4715-ab9f-7916d069a945


packages/widget/src/components/BottomSheet/BottomSheet.tsx
```diff
@@ -45,6 +53,8 @@ export const BottomSheet = forwardRef<BottomSheetBase, BottomSheetProps>(
         ModalProps={modalProps}
         slotProps={slotProps}
         disableAutoFocus
+        keepMounted={true}
+        inert={isInert}
```
chybisov, line 58: Can we use `drawerOpen` here to avoid having additional state + setTimeout?

effie-ms: The idea is that we need to apply inert **before** `drawerOpen` becomes false. That's why there is `isInert`. And if I update `setIsInert` without `setTimeout`, both `setDrawerOpen` and `setIsInert` are going to update in the same batch, not one after another, so it also does not work.

chybisov: Can we play with React transitions here? Maybe it can help https://react.dev/reference/react/startTransition
https://react.dev/reference/react/useTransition

effie-ms:
> During my test , after the wallet is connected, the wallet popover still has the focus trap warning.
yeah, reproduced it as well. clicks outside of dialogs are handled on MUI's side, so most probably we won't be able to do anything about them, as we are literally inserting "inert" right before closing.

PR #504, opened 2025-08-08
fix: scroll to index retry warnings

effie-ms:
https://lifi.atlassian.net/browse/LF-14991

`scrollToIndex` from Tanstack’s `useVirtualizer` issues warnings after 10 attempts to reach the desired offset, even though it works as expected visually. The likely cause is a small error in offset calculations that pushes the value slightly out of range.

This PR calculates the offset using Tanstack’s own function, but replaces the scrolling step with a standard `scrollTo` to bypass the built-in checks and retries — since we don’t care much about the small difference between the target offset and the actual offset.


packages/widget/src/components/Chains/VirtualizedChainList.tsx
```diff
@@ -78,6 +78,23 @@ export const VirtualizedChainList = ({
...
+          top: offsetInfo[0],
+          left: 0,
+          behavior: 'smooth',
+        })
+      })
+    },
+    [getOffsetForIndex, scrollElementRef.current]
+  )
+
   useLayoutEffect(() => {
```
chybisov: Do you think it makes sense to change `useLayoutEffect` to `useEffect`?

effie-ms: Not really, any win from this? The scroll should start earlier via`useLayoutEffect`, so in the end there will be less movement, I would leave it as is

chybisov: It's just designed for a bit different things. I don't think we need to scroll before repaint here. Wdyt? https://react.dev/reference/react/useLayoutEffect 😁

effie-ms: Okay, let's consider it is not urgent enough for `useLayoutEffect`, updated 😄

PR #513, opened 2025-08-15
feat: add meta/og tags to playgrounds

effie-ms:
https://lifi.atlassian.net/browse/LF-15107

Adds og, twitter and other meta tags to Vite and Next.js playgrounds.
**NB:** the image will be visible only after it ends up on the playground - SM will not see it via relative path.  

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,4 +9,4 @@ _If applicable, attach screenshots, GIFs, or videos to showcase the functionalit
 ## Checklist before requesting a review  
 - [ ] I have performed a self-review of my code.  
 - [ ] This pull request is focused and addresses a single problem.  
-- [ ] If this PR modifies the Widget API, I have updated the documentation (or submitted a change request on GitBook).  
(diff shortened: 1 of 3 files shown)


packages/widget-playground-next/src/app/layout.tsx
```diff
@@ -1,9 +1,48 @@
...
+const siteKeywords = [
+  'LI.FI',
+  'widget',
+  'DeFi',
+  'cross-chain',
+  'bridge',
+  'swap',
+  'playground',
+]
+const ogImage = '/og_image_playground.png'
```
yolodevz, line 17: about file system og image resolution - [link](https://nextjs.org/docs/app/api-reference/file-conventions/metadata/opengraph-image#image-files-jpg-png-gif)

packages/widget-playground-vite/index.html
```diff
@@ -19,7 +19,30 @@
       user's mobile device or desktop. See https://developers.google.com/web/fundamentals/web-app-manifest/
```
yolodevz, line 19: is this generated somehow, or does it require manual update here and in the `src/app` separately?

PR #515, opened 2025-08-15
feat: add checkbox customization

effie-ms:
https://lifi.atlassian.net/browse/LF-15063

- Replaces checkmark icons with checkboxes 
- Adds customization for MUI Checkbox
- Optimizes re-renders for Bridges and Exchanges

https://github.com/user-attachments/assets/f8375bd5-123b-4e9e-9b62-4b47862c6864

### DIFF
--- a/packages/widget/src/components/Tools/ToolItem.tsx
+++ b/packages/widget/src/components/Tools/ToolItem.tsx
@@ -1,5 +1,5 @@
 import { Avatar, Checkbox, ListItemAvatar } from '@mui/material'
-import { memo, useCallback } from 'react'
+import { memo } from 'react'
 import { ListItemText } from '../ListItemText.js'
 import { SettingsListItemButton } from '../SettingsListItemButton.js'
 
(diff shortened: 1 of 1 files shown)


packages/widget/src/components/Tools/ToolItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+      onToggle(tool.key)
+    }, [onToggle, tool.key])
+
+    const handleCheckboxClick = useCallback(
+      (e: React.MouseEvent) => {
+        e.stopPropagation()
+        onToggle(tool.key)
+      },
+      [onToggle, tool.key]
+    )
```
chybisov: In this case, useCallback doesn't really help with anything. It only works when you pass a function to a component, and we know that we wrap it in useCallback to prevent re-renders. Here we just use it directly as an onClick handler.

effie-ms: Updated

PR #519, opened 2025-08-19
feat: add an config option to force internal management

tomiiide:
https://lifi.atlassian.net/browse/LF-15091

Currently when an external context for an ecosystem is detected, e.g ETH, we reuse the external context and remove all ETH wallets from wallet management.

This config option provides a way to force the widget to keep it's internal context and wallet management even when an external context is detected.

https://github.com/user-attachments/assets/3cda2dee-67e1-4037-8ead-671d56eb8caa

- [x] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the public-docs repository. => https://github.com/lifinance/public-docs/pull/17

### DIFF
--- a/packages/widget-playground/src/components/DrawerControls/DesignControls/WalletManagementControl.tsx
+++ b/packages/widget-playground/src/components/DrawerControls/DesignControls/WalletManagementControl.tsx
@@ -75,7 +75,7 @@ export const WalletManagementControl = () => {
         </CardRowContainer>
         <CardRowContainer>
           <CardTitleContainer>
-            <CardValue>Force Internal Wallet Management</CardValue>
+            <CardValue>Force internal wallet management</CardValue>
           </CardTitleContainer>
           <Switch
             checked={isForceInternalWalletManagement}


packages/widget-playground/src/store/widgetConfig/useConfigValues.ts
```diff
@@ -117,8 +117,10 @@ export const useConfigWalletManagement = () => {
     : { onConnect: () => {} }
 
   return {
+    replacementWalletConfig,
     isExternalWalletManagement: !!walletConfig,
     isPartialWalletManagement: !!walletConfig?.usePartialWalletManagement,
-    replacementWalletConfig,
+    isForceInternalWalletMangement:
```
effie-ms: `isForceInternalWalletMangement` - could you update spelling here? :)

tomiiide: fixed, thanks.

effie-ms: there are still a few `isForceInternalWalletMangement` references in `WalletManagementControl` and `useConfigValues`

packages/widget-playground/src/components/DrawerControls/DesignControls/WalletManagementControl.tsx
```diff
@@ -61,6 +73,16 @@ export const WalletManagementControl = () => {
             aria-label="Use partial wallet management"
           />
         </CardRowContainer>
+        <CardRowContainer>
+          <CardTitleContainer>
+            <CardValue>Force Internal Wallet Management</CardValue>
```
chybisov: Let’s keep the same letter case for wallet management

Image

chybisov (changes requested):
Let's fix one small detail, and it should be good to go 😁

packages/widget-playground/src/components/DrawerControls/DesignControls/WalletManagementControl.tsx
tomiiide: Fixed!

PR #521, opened 2025-08-25
feat: prototype for modular architecture, sui example

effie-ms:
This is an example of how to isolate ecosystem-related packages not to bundle them within the widget when users do not use them. Draft, demoed on SUI.

The idea is that wallet context providers with their dependencies are packaged to separate packages (e.g. @lifi/wallet-provider-sui), and then, if they are installed and imported by users and passed as props to <LifiWidget />, the widget renders them. On render, all the handlers and info needed by wallet-management get set to the @lifi/wallet-store. And when wallet-management needs anything, t reads the info/handlers from the store (not exports from external libs as before). So both widget and wallet-management become independent from external dependencies. 
A separate package for the store is needed to avoid circular dependencies.

scheme


packages/widget-provider/src/types.ts
```diff
@@ -0,0 +1,52 @@
...
+  disconnect: () => Promise<void>
+  // EVM handlers
+  getBytecode?: (
+    chainId: number,
+    address: string
+  ) => Promise<string | undefined>
+  getTransactionCount?: (
+    chainId: number,
+    address: string
+  ) => Promise<number | undefined>
```
chybisov, line 50: Can we abstract ecosystem-specific methods so they don't pollute the types for other ecosystems?

packages/wallet-management/package.json
```diff
@@ -72,16 +66,13 @@
     "madge": "^8.0.0",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
-    "typescript": "^5.9.3"
+    "typescript": "^5.9.3",
+    "@lifi/widget-provider": "workspace:*"
```
chybisov: Should this be in `dependencies`?

packages/widget-provider-solana/package.json
```diff
@@ -0,0 +1,27 @@
+{
```
chybisov, line 1: We should extend this and other new package.json files with other attributes that are relevant for published packages, see widget or wallet management, for example.

packages/widget/src/types/widget.ts
```diff
@@ -395,7 +395,8 @@ export interface WidgetConfigPartialProps {
 export type WidgetProps = WidgetDrawerProps &
   WidgetConfig &
   WidgetConfigPartialProps &
-  FormRefProps
+  FormRefProps &
+  WidgetWalletProvidersProps
```
chybisov: What is the rationale for placing this in WidgetProps rather than WidgetConfig?

packages/widget/src/providers/WalletProvider/WalletProvider.tsx
```diff
@@ -2,27 +2,40 @@ import type { WalletManagementConfig } from '@lifi/wallet-management'
...
+          walletConfig?.forceInternalWalletManagement
+        }
+        chains={chains ?? []}
+      >
+        {WidgetWithProviders}
+      </ProviderComponent>
+    )
+  }
+
+  return WidgetWithProviders
```
chybisov: Maybe we can use reduceRight here to improve readability:

```ts
const EMPTY_CHAINS: Chain[] = []

export const WalletProvider = ({
  children,
  providers,
}: PropsWithChildren<WidgetWalletProvidersProps>) => {
  const { walletConfig } = useWidgetConfig()
  const { chains } = useAvailableChains()

  // Avoid recreating [] on every render
  const safeChains = chains ?? EMPTY_CHAINS

  // Initial provider block, extracted for clarity
  const base = (
    <>
      <SDKProviders />
      <WalletMenuProvider>{children}</WalletMenuProvider>
    </>
  )

  // Wrap providers from outermost -> innermost
  const wrapped = providers.reduceRight(
    (acc, Provider) => (
      <Provider
        forceInternalWalletManagement={walletConfig?.forceInternalWalletManagement}
        chains={safeChains}
      >
        {acc}
      </Provider>
    ),
    base
  )

  return wrapped
}
```

packages/widget/src/hooks/useAddressActivity.ts
```diff
@@ -1,33 +1,48 @@
...
 
 interface AddressActivity {
   hasActivity: boolean
   isLoading: boolean
   isFetched: boolean
   toAddress: string | undefined
 }
-
 export const useAddressActivity = (chainId?: number): AddressActivity => {
+  const { keyPrefix } = useWidgetConfig()
```
chybisov: Why do we need `keyPrefix` here?

packages/widget/src/hooks/useIsContractAddress.ts
```diff
@@ -12,23 +14,28 @@ export const useIsContractAddress = (
   isLoading: boolean
   isFetched: boolean
 } => {
+  const { keyPrefix } = useWidgetConfig()
```
chybisov: Why do we need `keyPrefix` here?

packages/widget/src/App.tsx
```diff
@@ -23,7 +23,11 @@ export const App = forwardRef<WidgetDrawer, WidgetProps>((props, ref) => {
 
   if (config.variant === 'drawer') {
     return (
-      <AppProvider config={config} formRef={props.formRef}>
+      <AppProvider
```
chybisov: We should also adjust these ones if we move providers to the usual config.

packages/widget/src/providers/WalletProvider/WalletProvider.tsx
```diff
@@ -2,27 +2,40 @@ import type { WalletManagementConfig } from '@lifi/wallet-management'
...
-          <SuiProvider>
-            <SDKProviders />
-            <WalletMenuProvider>{children}</WalletMenuProvider>
-          </SuiProvider>
-        </UTXOProvider>
-      </SVMProvider>
-    </EVMProvider>
+export const WalletProvider = ({
+  children,
+  providers,
```
chybisov, line 23: We need to think about how to ensure the providers array is memoized, because without referential stability, any parent re-render will recreate provider references and trigger a full remount of the widget provider tree 🤔

chybisov: Can we check how Dynamic handles memoization for this flow? They should already have something in place, and if not, this could introduce significant performance issues.

packages/widget-provider-solana/package.json
effie-ms: Extended 👌

packages/wallet-management/package.json
effie-ms: Yes! Thanks, updated 😅

packages/widget/src/hooks/useAddressActivity.ts
effie-ms: We add it to all the other `useQuery`'s query keys (`queryKey: [getQueryKey('tokens', keyPrefix)]`, `queryKey: [getQueryKey('routes', keyPrefix)]`, etc.), I guess they should be here as well, right?

packages/widget/src/hooks/useIsContractAddress.ts
effie-ms: Same as here: https://github.com/lifinance/widget/pull/521/files/465178fb5e2fc21069c7f4d6fac7883bcfd4bcdc#r2527505368

packages/widget/src/providers/WalletProvider/WalletProvider.tsx
effie-ms: Updated 🙏

packages/widget-provider/src/types.ts
effie-ms: Separated 👌

packages/widget/src/types/widget.ts
effie-ms: Currently we serialize `config` values to localStorage via zustand store, so we can't put components there. Unless we disassemble config to exclude provider components before passing them to the store. What do you think?

packages/widget/src/providers/WalletProvider/WalletProvider.tsx
effie-ms: Dynamic does not use providers in their implementation, they pass connector objects.

Maybe something like this: https://github.com/lifinance/widget/pull/521/commits/9fdcc5d2b470fab9aad065908f445444ef802d02 ?

packages/widget/src/App.tsx
effie-ms: https://github.com/lifinance/widget/pull/521#discussion_r2527579344 - let's decide 😄

packages/widget/src/types/widget.ts
chybisov: I believe this only happens on the playground, and we have some logic there to exclude certain props from serialization.

packages/widget/src/hooks/useAddressActivity.ts
chybisov: There are some queries that should have keyPrefix when there are two widgets on the page (Jumper does this), but for some of them it's okay to be shared, that's why those didn't have it.

packages/widget/src/providers/WalletProvider/WalletProvider.tsx
chybisov: That one can work for simple cases, but let's run some tests once this is merged (to make it easier). We should check how it behaves when we re-render the integrated widget with providers and how big a performance issue it is.

packages/widget/src/types/widget.ts
effie-ms: Moved `providers` to config 👌

packages/widget/src/App.tsx
effie-ms: Updated

packages/widget/src/hooks/useAddressActivity.ts
effie-ms: Removed `keyPrefix` from `useIsContractAddress` and `useAddressActivity`

PR #522, opened 2025-08-26
fix: do not show insufficient gas with auto refuel enabled

effie-ms:
https://lifi.atlassian.net/browse/LF-15118

Before (playground)
Screenshot 2025-08-25 at 22 17 29
After
Screenshot 2025-08-25 at 22 16 34

### DIFF
--- a/packages/widget/src/components/Messages/GasSufficiencyMessage.tsx
+++ b/packages/widget/src/components/Messages/GasSufficiencyMessage.tsx
@@ -3,6 +3,7 @@ import { type BoxProps, Typography } from '@mui/material'
 import { useTranslation } from 'react-i18next'
 import { formatUnits } from 'viem'
 import type { GasSufficiency } from '../../hooks/useGasSufficiency.js'
+import { useSettings } from '../../stores/settings/useSettings.js'
 import { AlertMessage } from './AlertMessage.js'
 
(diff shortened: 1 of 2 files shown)


packages/widget/src/components/Messages/GasSufficiencyMessage.tsx
```diff
@@ -14,6 +15,11 @@ export const GasSufficiencyMessage: React.FC<GasSufficiencyMessageProps> = ({
   ...props
 }) => {
   const { t } = useTranslation()
+  const { enabledAutoRefuel } = useSettings(['enabledAutoRefuel'])
+  // Do not show the message if gas auto refuel is enabled
+  if (enabledAutoRefuel) {
+    return null
+  }
```
chybisov: I think there is more to it. If the user doesn't have gas on the source chain, this will also hide the message, but we should show it in that case.

We probably need to change something inside the gas insufficiency message hook calculation.

chybisov (changes requested):

packages/widget/src/components/Messages/GasSufficiencyMessage.tsx
effie-ms: Updated the hook 👌

PR #534, opened 2025-09-02
fix: missing usd conversion

effie-ms:
When no wallet is connected, USD amount is 0:
Screenshot 2025-09-02 at 17 40 15

Added fallback to token with no balance.


chybisov (approved):
Thanks for finding this! 🙂

PR #544, opened 2025-09-09
fix: return from queryFn

effie-ms:
image

Addresses the undefined return. The filtering logic should already work fine.

### DIFF
--- a/packages/widget/src/hooks/useTokenSearch.ts
+++ b/packages/widget/src/hooks/useTokenSearch.ts
@@ -46,7 +46,7 @@ export const useTokenSearch = (
             t.address.toLowerCase()
           )
         ) {
-          return undefined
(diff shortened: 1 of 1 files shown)


packages/widget/src/hooks/useTokenSearch.ts
```diff
@@ -75,7 +75,7 @@ export const useTokenSearch = (
     retry: false,
   })
   return {
-    token: data,
+    token: data || undefined,
```
chybisov, line 78: Can we just return null here? Not a big deal, I think 🤔

effie-ms: We do have a few components that will require overwriting `token` type to null then. Possible, but not sure if worth it 😅

PR #545, opened 2025-09-09
feat: dynamic loading for translations

effie-ms:
https://lifi.atlassian.net/browse/LF-15428

This PR implements dynamic json translations loading (with import()) for widget and wallet management packages, reducing the initial bundle size by loading translations on-demand instead of preloading all languages.  

**Widget:**
On creation of settingsStore (first load):
- Determine initial language (last cached or `widgetConfig.languages.default`).
- Obtain initial language translations (last cached or dynamically loaded from the corresponding file).

In I18nProvider, initialize i18n instance:
- Use initial language and its translations for i18n instance initialization.
- Use English language and statically imported English translations as fallback.
- Custom `widgetConfig.languageResources` of new (not existing in our files) languages are added statically.

When language changes:
- Dynamically load a new language resource and add it to i18n instance.
- Cache it (if the page reloads, we have sooner access to the translations - to avoid blinking).

**Wallet management:**
- Receives language from widget's i18n instance
- Statically imports enResource from `en.json` (fallback). 
- Dynamically imports other language resources.

**Testing considerations:**
- Widget can receive language configuration (`languages` and `languageResources`) from widget config.
- When `widgetConfig.languages.default` is set, it should be loaded first (incognito mode or clearing site data, test by setting `widgetConfig.languages.default` to non-English keys)
- `widgetConfig.languageResources` can extend existing languages and/or introduce new languages. When a new language is introduced, it should show up in the languages selector, it should be possible to set it as `widgetConfig.languages.default` and it should work similarly to existing language resources ([docs](https://docs.li.fi/widget/internationalization#configure-languages)).

This PR is mostly performance optimization. For users/integrators, the only change is English fallback (always, even if English is not allowed), and the rest is expected to work similarly to what it was before).

### DIFF
--- a/packages/widget/src/components/Header/NavigationHeader.tsx
+++ b/packages/widget/src/components/Header/NavigationHeader.tsx
@@ -22,7 +22,10 @@ export const NavigationHeader: React.FC = () => {
     useWidgetConfig()
   const { navigateBack } = useNavigateBack()
   const { account } = useAccount()
-  const { element, title } = useHeaderStore((state) => state)
+  const [element, title] = useHeaderStore((state) => [
+    state.element,
+    state.title,
+  ])
   const { pathname } = useLocation()
 
   const cleanedPathname = pathname.endsWith('/')
--- a/packages/widget/src/hooks/useHeader.ts
+++ b/packages/widget/src/hooks/useHeader.ts
@@ -3,7 +3,10 @@ import { useEffect } from 'react'
 import { useHeaderStore } from '../stores/header/useHeaderStore.js'
 
 export function useHeader(title: string, action?: ReactNode) {
-  const { setTitle, setAction } = useHeaderStore((state) => state)
+  const [setTitle, setAction] = useHeaderStore((state) => [
+    state.setTitle,
+    state.setAction,
+  ])
 
   useEffect(() => {
     const removeTitle = setTitle(title)


packages/widget/src/providers/I18nProvider/I18nProvider.tsx
```diff
@@ -75,7 +40,62 @@ export const I18nProvider: React.FC<React.PropsWithChildren> = ({
...
+        await loadLocale('en', languageResources?.en).then(
+          (languageResource) => {
+            i18nInstance.addResourceBundle(
+              'en',
+              'translation',
+              languageResource,
+              true,
+              true
+            )
+          }
```
chybisov: I think English should be statically imported, this way it is always available, and we have less delay

packages/widget/src/hooks/useTools.ts
```diff
@@ -1,12 +1,14 @@
...
 import { useQuery } from '@tanstack/react-query'
 import { useWidgetConfig } from '../providers/WidgetProvider/WidgetProvider.js'
-import { settingsStore } from '../stores/settings/useSettingsStore.js'
+import { useSettingsStore } from '../stores/settings/SettingsStore.js'
 import { getConfigItemSets, isItemAllowedForSets } from '../utils/item.js'
 import { getQueryKey } from '../utils/queries.js'
 
 export const useTools = () => {
   const { bridges, exchanges, keyPrefix } = useWidgetConfig()
+  const initializeTools = useSettingsStore((state) => state.initializeTools)
```
chybisov: Previously, this was accessed imperatively via getState() to avoid re-renders. Can we keep the same approach with the new store?

effie-ms: Reverted

packages/widget/src/stores/settings/createSettingsStore.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  // Initialize language settings on store creation
+  const initializeLanguageSettings = async () => {
+    const state = store.getState()
+    const initialLanguage =
+      state.getValue('language') || config?.languages?.default
+
+    // Preload translations (from existing translation files)
+    const customLanguages = Object.keys(config?.languageResources || {}).filter(
+      (key) => !allLanguages.includes(key as LanguageKey)
```
chybisov: Should we deep-merge custom resources with the loaded ones to preserve the overrides? I think I saw that before, but now it looks like we removed that 👀

packages/widget/src/i18n/pt.json
```diff
@@ -327,7 +327,7 @@
       "slow": "Lento",
       "title": "Preço do gás"
     },
-    "routePriority": "Prioridade de rota / caminho",
```
chybisov, line 330: We should change this in Crowdin, only English should be changed in the code https://crowdin.com/project/lifi-widget

packages/widget/src/providers/I18nProvider/I18nProvider.tsx
```diff
@@ -1,67 +1,77 @@
...
-        }
-        return resources
-      }, resources)
+  const i18nInstanceRef = useRef<i18n | null>(null)
+  const isWaitingForLanguageCache = useRef(false)
+  const i18nInstance = useMemo(() => {
+    // Allow reinitializing i18nInstance if languageCache was not set before
+    // (useful for very first load)
+    if (i18nInstanceRef.current && !isWaitingForLanguageCache.current) {
+      return i18nInstanceRef.current
```
chybisov: Are we sure we need to re-initialize it on languageCache change, maybe we can update the already existing instance inside useEffect? 

We should also make sure whenever languageCache changes, we update it, even when it's available at the first render, because it might slightly change when we load translations and set it again. Do you think this makes sense? 👀

packages/widget/src/stores/settings/createSettingsStore.ts
effie-ms: So these custom ones are unique, `allLanguages` has a list of keys corresponding to files, so in these lines, it takes all the other keys (for which we do not have files), and then there is nothing to merge it with :)

packages/widget/src/i18n/pt.json
effie-ms: Reverted

packages/widget/src/providers/I18nProvider/I18nProvider.tsx
effie-ms: UPD: no reinitialization now.
Added updating resource bundles to i18Instance's `useMemo` on `languageCache` change.

PR #546, opened 2025-09-10
fix: add form type to isAllNetworks

effie-ms:
https://lifi.atlassian.net/browse/LF-15459

Use this config: `chains: { to: { allow: [57073] } }`

1) `isAllNetworks` is a single flag right now, but we should actually have 2 values - for "fromChain" and "toChain" - added 2 flags to the store. No migration required as both have new variable names.
2) Right now, all networks shows unfiltered (by allow-deny) lists of tokens - fixed, see changes in `packages/widget/src/hooks/useTokenBalances.ts`. 
3) Since `isAllNetworks` flags are enabled by default, we should handle cases when there is only 1 chain in the list (we do not show "All networks" there, so we need to handle it as false). 

https://github.com/user-attachments/assets/38aebbbb-8d43-41b9-abad-36f106a8f724


packages/widget/src/hooks/useIsMultipleNetworks.ts
```diff
@@ -0,0 +1,15 @@
...
+import type { FormType } from '../stores/form/types.js'
+
+/** This hook returns isAllNetworks value per formType if there are multiple networks.
+ * Enabled isAllNetworks for a single available for selection network returns false. */
+export const useIsMultipleNetworks = (formType: FormType) => {
+  const { chains } = useChainSelect(formType)
+
+  const { getIsAllNetworks } = useChainOrderStore((state) => ({
+    getIsAllNetworks: state.getIsAllNetworks,
+  }))
```
chybisov: Can we do it like this?

```ts
const isAllNetworks = useChainOrderStore((state) => state.getIsAllNetworks(formType))
```

packages/widget/src/components/Chains/VirtualizedChainList.tsx
```diff
@@ -38,13 +41,17 @@ export const VirtualizedChainList = ({
...
       state.pinnedChains,
       state.setPinnedChain,
-      state.isAllNetworks,
       state.setIsAllNetworks,
-    ])
+    ]
+  )
+  const isAllNetworks = useChainOrderStore(
+    (state) => state[`${formType}IsAllNetworks`]
+  )
```
chybisov: We should put this inside the array above instead of calling store one more time 🙂

packages/widget/src/components/ChainSelect/ChainSelect.tsx
```diff
@@ -41,10 +41,12 @@ export const ChainSelect = memo(({ formType }: FormTypeProps) => {
...
-  const { isAllNetworks, setIsAllNetworks } = useChainOrderStore((state) => ({
-    isAllNetworks: state.isAllNetworks,
-    setIsAllNetworks: state.setIsAllNetworks,
-  }))
+  const { showAllNetworks, isAllNetworks, setIsAllNetworks } =
+    useChainOrderStore((state) => ({
+      showAllNetworks: state[`${formType}ShowAllNetworks`],
+      isAllNetworks: state[`${formType}IsAllNetworks`],
+      setIsAllNetworks: state.setIsAllNetworks,
+    }))
```
chybisov: Please use an array here instead of object, it's usually more efficient and the variable names can stay the same 🙏

effie-ms: Updated 👌

chybisov (approved):

PR #560, opened 2025-09-18
update reown example

tomiiide:
https://lifi.atlassian.net/browse/LF-15541

### What was done?
**The reown example was updated to be cleaner** 
- Move all appkit configs to the config file
- Separate Solana and EVM provider management logic 
- Stop blocking widget render until chains are loaded
- load chains dynamically, and update the appKit config with new chains using the `addNetwork` method.

**SyncWagmiConfig update**
We now preserve connectors that were being removed when syncing wagmi configs.

- [x] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the public-docs repository.


packages/wallet-management/src/syncWagmiConfig.ts
```diff
@@ -13,13 +13,26 @@ export const syncWagmiConfig = async (
...
+    const preservedConnectors = currentConnectors.filter((connector) => {
+      return !newMipdRdnsSet.has(connector.id)
+    })
+
+    const newConnectors = [
       ...connectors,
       ...(wagmiConfig._internal.mipd
         ?.getProviders()
         .map(wagmiConfig._internal.connectors.providerDetailToConnector) ?? []),
     ].map(wagmiConfig._internal.connectors.setup)
```
chybisov: I think we can reuse `mipdProviders` instead of getting them a second time here.

tomiiide: thanks, fixed

packages/wallet-management/src/syncWagmiConfig.ts
```diff
@@ -13,13 +19,27 @@ export const syncWagmiConfig = async (
...
+      mipdProviders.map((provider) => provider.info.rdns)
+    )
+
+    // Keep existing connectors that don't conflict with newly discovered MIPD providers
+    const preservedConnectors = currentConnectors.filter((connector) => {
+      return !newMipdRdnsSet.has(connector.id)
+    })
+
+    const newConnectors = [
       ...connectors,
```
chybisov: Should we also check these `connectors` when deciding which ones to preserve?
It seems possible that `preservedConnectors` could already contain the same connector as those provided in `connectors`. Is that the case, can we please check? 👀

examples/reown/src/providers/WalletProvider.tsx
```diff
@@ -1,93 +1,43 @@
...
-    wagmi.current = wagmiAdapter
-    modal.current = appKit
-  }
+  const { chains } = useAvailableChains()
 
-  const { wagmiConfig } = wagmi.current
+  const { isConnected, status } = useAppKitAccount()
+  const { isConnected: ethIsConnected } = useAppKitAccount({
+    namespace: 'eip155',
+  })
```
chybisov, line 22: Can we combine these two so we call it only once?

tomiiide: No, we can't. The second call is namespaced. You can think of them as two different functions, as the values are different.

We need to be able to compare the global `isConnected` with the eip `isConnected` to determine when to manually set the global state.

PR #573, opened 2025-10-03
refactor: use zustand create instead of createWithEqualityFn

effie-ms:
Updated outdated zustand store setups where we use `createWithEqualityFn` function with `Object.is` equality function, since this is the same default as a standard `create` function.

Added memoization to `useChainOrder` as it triggered infinite loop with the state changes.

(settingsStore is updated in https://github.com/lifinance/widget/pull/545, skipped not to case merge conflicts)

### DIFF
--- a/packages/widget-playground/src/store/editTools/createEditToolsStore.ts
+++ b/packages/widget-playground/src/store/editTools/createEditToolsStore.ts
@@ -1,13 +1,13 @@
 import type { WidgetTheme } from '@lifi/widget'
 import type { StateCreator } from 'zustand'
+import { create } from 'zustand'
 import { persist } from 'zustand/middleware'
-import { createWithEqualityFn } from 'zustand/traditional'
 import type { FormValues } from '../types.js'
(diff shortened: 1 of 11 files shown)


packages/widget-playground/src/store/editTools/createEditToolsStore.ts
```diff
@@ -189,6 +189,5 @@ export const createEditToolsStore = (initialTheme?: WidgetTheme) =>
           }
         },
       }
-    ) as StateCreator<ToolsState, [], [], ToolsState>,
-    Object.is
+    ) as StateCreator<ToolsState, [], [], ToolsState>
```
chybisov: Can we please check how we can remove these type casts?

chybisov (changes requested):

packages/widget-playground/src/store/editTools/createEditToolsStore.ts
effie-ms: Removed

PR #585, opened 2025-10-29
feat: apply client for modular SDK

effie-ms:
https://lifi.atlassian.net/browse/LF-15705 

Run with https://github.com/lifinance/sdk/pull/318

### DIFF
--- a/packages/widget-provider-bitcoin/src/providers/BitcoinProviderValues.tsx
+++ b/packages/widget-provider-bitcoin/src/providers/BitcoinProviderValues.tsx
@@ -5,7 +5,6 @@ import {
   getAccount,
   getConnectorClient as getBigmiConnectorClient,
 } from '@bigmi/client'
-import { isUTXOAddress } from '@bigmi/core'
 import { useAccount, useConfig, useConnect } from '@bigmi/react'
 import { ChainId, ChainType } from '@lifi/sdk'
 import { BitcoinProvider as BitcoinSDKProvider } from '@lifi/sdk-provider-bitcoin'
@@ -87,7 +86,6 @@ export const BitcoinProviderValues: FC<
         isExternalContext,
         connect: handleConnect,
         disconnect: handleDisconnect,
-        isValidAddress: isUTXOAddress,
       }}
     >
       {children}
--- a/packages/widget-provider-ethereum/src/providers/EthereumProviderValues.tsx
+++ b/packages/widget-provider-ethereum/src/providers/EthereumProviderValues.tsx
@@ -14,8 +14,9 @@ import {
   useMemo,
   useState,
 } from 'react'
-import { type Address, isAddress as isEVMAddress } from 'viem'
-import { type Connector, useAccount, useConfig, useConnect } from 'wagmi'
+import type { Address } from 'viem'
+import type { Connector } from 'wagmi'
+import { useAccount, useConfig, useConnect } from 'wagmi'
 import {
   connect,
   disconnect,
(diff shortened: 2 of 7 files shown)


packages/widget-playground/src/defaultWidgetConfig.ts
```diff
@@ -69,9 +69,9 @@ export const widgetBaseConfig: WidgetConfig = {
   // },
   sdkConfig: {
     apiUrl: 'https://li.quest/v1',
-    rpcUrls: {
-      [ChainId.SOL]: ['https://dacey-pp61jd-fast-mainnet.helius-rpc.com/'],
-    },
+    // rpcUrls: {
```
effie-ms: I have messed up the SDK logic a bit :smile:

In this line:
https://github.com/lifinance/sdk/blob/ad59f86579a2a51007b14aacaf39e33eeee00801/packages/sdk/src/client/getClientStorage.ts#L33
if we pass `config.rpcUrls` via config, initial _rpcUrls will not be empty, so `!Object.keys(_rpcUrls).length` will not make sense. this will need to be fixed, otherwise we don't get balances fetched as there won't be RPC URLs. 

Commenting this config line for now, but it should be reverted.

packages/widget-playground-vite/vite.config.ts
```diff
@@ -19,6 +19,7 @@ export default defineConfig({
     sourcemap: true,
   },
   optimizeDeps: {
+    exclude: ['@safe-global/safe-apps-sdk'],
```
effie-ms: Reown appkit uses this dependency, and that dependency requires `viem` that we removed. So it breaks the build. This might require some different fix - this is just a hotfix.

packages/widget-playground/src/defaultWidgetConfig.ts
effie-ms: Added fix here: https://github.com/lifinance/sdk/pull/328

packages/widget-playground-vite/vite.config.ts
effie-ms: Note: External wallet management is affected/not working as well.

chybisov: We can bring viem back, is it a peer dependency for those?

knip.json
```diff
@@ -1,4 +1,4 @@
 {
   "ignore": ["examples/**", "packages/widget-embedded/**", "scripts/**"],
-  "ignoreDependencies": ["@mui/system", "csstype", "@solana/web3.js"]
+  "ignoreDependencies": ["@mui/system", "csstype", "viem"]
```
effie-ms: "viem" dependency is needed for Vite playground for Reown's appkit, detected as unused otherwise.

packages/widget/src/index.ts
```diff
@@ -9,7 +9,6 @@ export * from './components/ContractComponent/NFT/types.js'
 export * from './components/Skeleton/WidgetSkeleton.js'
 export { defaultMaxHeight } from './config/constants.js'
 export * from './config/version.js'
-export { useAvailableChains } from './hooks/useAvailableChains.js'
```
effie-ms, line 12: `useAvailableChains` depends on SDK client now and cannot be used outside of the widget.

packages/widget-playground-vite/vite.config.ts
effie-ms: A normal dependency, put it back as it's just on the playground level.

packages/widget/src/index.ts
chybisov: Let's keep this in mind when updating examples and migration guide 👌

PR #587, opened 2025-10-29
feat: add tanstack-router

effie-ms:
https://lifi.atlassian.net/browse/LF-5228

Replaces `react-router-dom` with a memory router from`@tanstack/react-router` for nested in-memory routing.
1-to-1 replacement, routes stay the same. Solves the problem of conflicting multiple React routers.


examples/react-router-7/app/components/LiFiWidget.tsx
```diff
@@ -19,11 +14,11 @@ export function LiFiWidget() {
...
   return (
-    <Suspense fallback={<WidgetSkeleton config={config} />}>
-      <LiFiWidgetLazy
+    <ClientOnly fallback={<WidgetSkeleton config={config} />}>
+      <LiFiWidgetComponent
         config={config}
         integrator="remix-to-react-router-7-example"
       />
-    </Suspense>
+    </ClientOnly>
```
chybisov, line 22: Suspense no longer works here?

examples/tanstack-router/README.md
```diff
@@ -0,0 +1,42 @@
+# TanStack Router + LiFi Widget Example
```
chybisov: Please use `LI.FI Widget` brand name everywhere 🙏😄

packages/widget/src/hooks/useNavigateBack.ts
```diff
@@ -1,35 +1,17 @@
...
-      //     { replace: true },
-      //   );
-      // }
+  const navigateBack = useCallback(() => {
+    if (router.history.length > 1) {
+      router.history.go(-1)
+    } else {
+      router.navigate({ to: navigationRoutes.home, replace: true })
+    }
+  }, [router])
```
chybisov, line 14: Are we sure there is no scenario in which we want to navigate back to the previous page from some long path?

chybisov: There is at least one actually 😅

https://github.com/user-attachments/assets/63d813c0-cf18-404e-8331-a11acc79da3a

packages/widget/src/components/TokenList/TokenList.tsx
```diff
@@ -12,69 +14,73 @@ import type { TokenListProps } from './types.js'
...
-export const TokenList: FC<TokenListProps> = memo(
-  ({ formType, parentRef, height, onClick }) => {
-    const emitter = useWidgetEvents()
+export const TokenList: FC<TokenListProps> = memo(({ formType, headerRef }) => {
+  const navigateBack = useNavigateBack()
+  const listParentRef = useRef<HTMLUListElement | null>(null)
+  const { listHeight } = useListHeight({
+    listParentRef,
+    headerRef,
+  })
```
chybisov, line 23: I'm not sure why the height calculation was inside WrappedTokenList, but have you checked that it doesn't affect performance if we move it here? 👀 Also, just generally curious if that was applied in relation to changing the routing to tanstack.

packages/widget/src/pages/ActiveTransactionsPage/ActiveTransactionsPage.tsx
```diff
@@ -35,6 +37,16 @@ const DeleteIconButton: React.FC<IconButtonProps> = ({ onClick }) => {
...
 export const ActiveTransactionsPage = () => {
+  const { pathname } = useLocation()
+
+  if (pathname.endsWith(navigationRoutes.activeTransactions)) {
+    return <ActiveTransactionsPageComponent />
+  }
+
+  return <Outlet />
+}
+
```
chybisov: Why is this wrapper necessary?

packages/widget/src/pages/SelectTokenPage/SelectTokenPage.tsx
```diff
@@ -1,22 +1,36 @@
...
+    return <SelectTokenComponent formType="from" />
+  }
+
+  if (pathname.endsWith(navigationRoutes.toToken)) {
+    return <SelectTokenComponent formType="to" />
+  }
+
+  return <Outlet />
+}
+
```
chybisov: I believe there is a way we can just put <SelectTokenComponent formType="from" /> and <SelectTokenComponent formType="to" /> to the `component` property when building routes, so we don't have such checks. Like

```ts
const someRoute = createRoute({
  getParentRoute: () => ...,
  path: ...,
  component: SelectTokenFromComponent,
})

function SelectTokenFromComponent() {
  return <SelectTokenComponent formType="from" /> 
}
```

packages/widget/src/components/Routes/RoutesExpanded.tsx
```diff
@@ -60,8 +60,9 @@ export const RoutesExpanded = memo(function RoutesExpanded({
   const onRouteClick = useCallback(
     (route: Route) => {
       setReviewableRoute(route)
-      navigate(navigationRoutes.transactionExecution, {
-        state: { routeId: route.id },
+      navigate({
+        to: navigationRoutes.transactionExecution,
+        search: { routeId: route.id },
```
chybisov, line 65: Would it be better to use this concept here instead of a search? 👀 https://tanstack.com/router/latest/docs/framework/react/routing/code-based-routing#dynamic-route-segments

packages/widget/src/pages/SendToWallet/SendToWalletPage.tsx
```diff
@@ -38,7 +38,17 @@ import {
...
-export const SendToWalletPage = () => {
+export const SendToWalletPage: React.FC = () => {
+  const { pathname } = useLocation()
+
+  if (pathname.endsWith(navigationRoutes.sendToWallet)) {
+    return <SendToWalletComponent />
+  }
+
+  return <Outlet />
+}
```
chybisov: We should find a way to remove these wrappers 😅

packages/widget/src/AppDefault.tsx
```diff
@@ -1,35 +1,225 @@
...
+    const history = createMemoryHistory({
+      initialEntries: ['/'],
+    })
+
+    return createRouter({
+      routeTree,
+      history,
+      defaultPreload: 'intent',
+    })
+  }, [])
```
chybisov: Does it make sense to have it wrapped in useMemo, or can we move it out of the component like other creation functions above?

examples/react-router-7/app/components/LiFiWidget.tsx
effie-ms: Yes, it throws some hydration warning on console, so I reused our `ClientOnly` wrapper

examples/tanstack-router/README.md
effie-ms: Updated the brand name in the example 👍

packages/widget/src/components/TokenList/TokenList.tsx
effie-ms: so `WrappedTokenList` is just like this, so I removed completely and moved the listHeight hook to `TokenList`, not sure it should change anything performance-wise

```
const WrappedTokenList = ({ headerRef, formType }: WrappedTokenListProps) => {
  const { navigateBack } = useNavigateBack()
  const listParentRef = useRef<HTMLUListElement | null>(null)
  const { listHeight } = useListHeight({
    listParentRef,
    headerRef,
  })
  return (
    <TokenList
      parentRef={listParentRef}
      height={listHeight}
      onClick={navigateBack}
      formType={formType}
    />
  )
}
```

packages/widget/src/AppDefault.tsx
effie-ms: We can move it out, updated 👌

packages/widget/src/components/Routes/RoutesExpanded.tsx
effie-ms: So there are `params` and `search`, and `params` are dynamic values (ids, hashes, etc.) - we don't really use any paths like this, only static values. Do you see any use cases for them?

packages/widget/src/pages/SendToWallet/SendToWalletPage.tsx
effie-ms: I have found one way 😄  
We can create one more route returning `<Outlet />` as a component, and the other - returning the component itself by the relative path "/". Probably cleaner. Unfortunately we must have `<Outlet />` somewhere, otherwise nested pages will not render.

packages/widget/src/pages/SelectTokenPage/SelectTokenPage.tsx
effie-ms: Updated, related to https://github.com/lifinance/widget/pull/587#discussion_r2534757573

packages/widget/src/pages/ActiveTransactionsPage/ActiveTransactionsPage.tsx
effie-ms: Updated, related to https://github.com/lifinance/widget/pull/587#discussion_r2534757573

packages/widget/src/components/TokenList/TokenList.tsx
chybisov: There might have been some issues with the blank token list page or something, because useListHeight triggers additional render, and together with the virtualized list, there was possibly an issue. It might be resolved tho, can you please test the token list thoroughly to make sure nothing breaks due to this change? 🙏😅

packages/widget/src/components/Routes/RoutesExpanded.tsx
chybisov: I don't have a strong opinion on this, but we have route id and params might be a better option now, since we have a truly in-memory router. Previously, it was done via state + search to overcome the limitations of the browser router 😄

effie-ms: If we use `params`, the path will look like `/transaction-execution/{hash}`, do we want to change it? And the emitted events on path change (we send pathnames to Jumper) will have to be updated then as well. Should we?

packages/widget/src/components/TokenList/TokenList.tsx
effie-ms: I haven't managed to get that issue with the virtualized list, maybe it was related to something else.

packages/widget/src/hooks/useNavigateBack.ts
effie-ms: Resolved with back actions: https://github.com/lifinance/widget/pull/587/commits/ec919e71d5d00d3aa5274e954e3ec4c4cdba7258

packages/widget/src/components/Routes/RoutesExpanded.tsx
chybisov: Ah, I see, then let's not do it ☺️ Thanks for looking into it 😄

chybisov (approved):

Issue #583, opened 2025-10-27
LiFiWidget overrides global <html> theme class and causes flicker in Next.js with next-themes


soloviovqqq:
When integrating the @lifi/widget into a Next.js application that uses next-themes for global dark/light theme handling, the widget automatically modifies the <html> element on mount.

Specifically, it adds or replaces theme-related classes (such as "light" or "dark") and applies color-scheme styles globally.
This interferes with the app’s existing theme system and results in visual flickering (the page briefly flashes white or changes to the wrong theme).

chybisov:
@soloviovqqq hey, could you please provide more details on what the integration looks like (configuration, project setup, Next.js version, UI library used in the app, etc.) or share a reproducible example?

soloviovqqq:
Reproducible example: https://github.com/soloviovqqq/lifi-example

Steps to reproduce
• Open LI Fl drawer (click on the button below)
• In LI Fl drawer settings change appearance to dark
• Close LI Fl drawer
Set the nextjs theme to light
• Refresh the page
• When the component <Widget/> is mounted, the <html> tag is modified (if comment component < Widget/> everything works properly)


Expected behavior
The widget should not modify the html element or global theme classes (dark, light, etc.). It should only apply its theme internally based on the provided

Important note
You dont need to connect any wallet to reproduce this issue.


https://github.com/user-attachments/assets/91831cd5-7b69-4ae0-8ec4-b74829682718

PR #589, opened 2025-10-31
fix: prevent overriding global theme from widget

effie-ms:
https://lifi.atlassian.net/browse/LF-16412

_UPD:
Using the example provided in https://github.com/lifinance/widget/issues/583, just removing `setMode()` will not be sufficient.
The problem is that MUI's mode switcher (`setMode()`) always overwrites global html attributes, no matter what properties are set when defining the theme: when we define` colorSchemeSelector: 'class'` (or any other attribute) in createTheme (from MUI), it is always attached to the `html` node. 

MUI's `ThemeProvider` has a property `colorSchemeNode` which allows to attach the color scheme attribute to a certain node (not only to global `html`), which theoretically could scope the mode only within widget's container, but instead it just duplicates the attributes, setting them to both global html node and that specified node.

So setting the node in this form still does not resolve the problem.
```
<div id="lifi-widget-custom-root">
      <MuiThemeProvider
        ...
        colorSchemeNode={document.getElementById('lifi-widget-custom-root') as Element}
      >
        {children}
      </MuiThemeProvider>
</div>
```
https://github.com/user-attachments/assets/53dbd8c8-1be9-40db-a792-1ae865d962e5

**_Instead, conditionally attaching widget to shadow root:_** https://mui.com/material-ui/customization/shadow-dom/

- [ ] I have performed a self-review and testing of my code.  

### DIFF
--- a/packages/widget-playground/src/components/Widget/WidgetView.tsx
+++ b/packages/widget-playground/src/components/Widget/WidgetView.tsx
@@ -32,15 +32,13 @@ export function WidgetView() {
   return (
     <WidgetViewContainer toggleDrawer={toggleDrawer}>
       {!isSkeletonShown || isSkeletonSideBySide ? (
-        <div style={{ margin: 'auto' }}>
-          <LiFiWidget
-            config={config}
-            ref={drawerRef}
-            integrator="li.fi-playground"
-            formRef={formRef}
-            open
-          />
-        </div>
+        <LiFiWidget
+          config={config}
+          ref={drawerRef}
+          integrator="li.fi-playground"
+          formRef={formRef}
+          open
+        />
       ) : null}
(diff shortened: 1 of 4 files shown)


effie-ms:
We have observed performance issues with wrapping the widget into shadow root, so closing it for now.

Decided to proceed with already existing workaround: setting the theme via config: hiding the theme switcher in the widget UI (`HiddenUI.Appearance`) and passing the theme mode into config (`appearance: 'light' | 'dark' | 'system'`). Then the theme can be switched externally, and the widget will just adapt to the change. 

Docs:
https://docs.li.fi/widget/customize-widget#appearance
https://docs.li.fi/widget/customize-widget#hidden-ui-elements

PR #594, opened 2025-11-12
feat: add setting to hide small balances

effie-ms:
https://lifi.atlassian.net/browse/LF-16062 

https://github.com/user-attachments/assets/f9e87389-d508-48f5-a83b-25ceb5f44a31

### DIFF
--- a/packages/widget/src/hooks/useTokenBalances.ts
+++ b/packages/widget/src/hooks/useTokenBalances.ts
@@ -100,7 +100,7 @@ export const useTokenBalances = (
       !!smallBalanceThreshold && !hiddenUI?.includes(HiddenUI.HideSmallBalances)
     const threshold = hideSmallBalances
       ? Number.parseFloat(smallBalanceThreshold)
-      : NaN
+      : undefined
 
     if (!balancesByChain) {
       return undefined
@@ -122,10 +122,9 @@ export const useTokenBalances = (
       let processedToken = token
       if (
         hideSmallBalances &&
-        !Number.isNaN(threshold) &&
+        threshold !== undefined &&
         threshold >= 0 &&
-        token.amount &&
-        token.amount !== 0n
+        token.amount
       ) {
         const balanceUSD = formatTokenPrice(
           token.amount,
--- a/packages/widget/src/pages/SettingsPage/SettingsCard/SettingsAccordian.tsx
+++ b/packages/widget/src/pages/SettingsPage/SettingsCard/SettingsAccordian.tsx
@@ -24,7 +24,7 @@ export const useSettingsCardExpandable = () => {
   const { openCard, setOpenCard } = useContext(SettingsAccordionContext)
 
   const toggleExpanded = (forceExpanded?: boolean) => {
-    const newExpanded = forceExpanded !== undefined ? forceExpanded : !expanded
+    const newExpanded = forceExpanded ?? !expanded
     setExpanded(newExpanded)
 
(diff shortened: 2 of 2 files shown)


packages/widget/src/hooks/useTokenBalances.ts
```diff
@@ -53,20 +56,48 @@ export const useTokenBalances = (
...
+          if (balanceUSD < threshold) {
+            return {
+              ...token,
+              amount: 0n,
+            }
+          }
+          return token
+        })
+      }
+    }
```
chybisov: Is there an opportunity to use this filter somewhere we are already mapping these tokens and their balances to avoid an additional heavy loop?

effie-ms: Made a reduce() with the filter above, should be 1 pass less 👌

packages/widget/src/pages/SettingsPage/SettingsCard/SettingsAccordian.tsx
```diff
@@ -22,8 +22,9 @@ export const useSettingsCardExpandable = () => {
   const settingCardExpandableId = useId()
   const [expanded, setExpanded] = useState(false)
   const { openCard, setOpenCard } = useContext(SettingsAccordionContext)
-  const toggleExpanded = () => {
-    const newExpanded = !expanded
+
+  const toggleExpanded = (forceExpanded?: boolean) => {
+    const newExpanded = forceExpanded !== undefined ? forceExpanded : !expanded
```
chybisov: Can we simplify it to `const newExpanded = forceExpanded ?? !expanded`?

packages/widget/src/hooks/useTokenBalances.ts
```diff
@@ -53,20 +58,62 @@ export const useTokenBalances = (
...
+      }
+
+      // Apply small balance threshold transformation if enabled
+      let processedToken = token
+      if (
+        hideSmallBalances &&
+        !Number.isNaN(threshold) &&
+        threshold >= 0 &&
+        token.amount &&
+        token.amount !== 0n
```
chybisov: `token.amount !== 0n` is kinda redundant here

packages/widget/src/hooks/useTokenBalances.ts
```diff
@@ -53,20 +58,62 @@ export const useTokenBalances = (
         (t) => `${t.chainId}-${t.address.toLowerCase()}`
       ) || []
     )
-    return balancesByChain?.filter((token) => {
+
+    const hideSmallBalances =
+      !!smallBalanceThreshold && !hiddenUI?.includes(HiddenUI.HideSmallBalances)
+    const threshold = hideSmallBalances
+      ? Number.parseFloat(smallBalanceThreshold)
+      : NaN
```
chybisov: Are we sure we want to assign `NaN` to the threshold here? 😅

effie-ms: can be undefined instead, updated

packages/widget/src/hooks/useTokenBalances.ts
effie-ms: indeed, thanks, updated 👌

packages/widget/src/pages/SettingsPage/SettingsCard/SettingsAccordian.tsx
effie-ms: yes, updated 👍

PR #600, opened 2025-11-19
chore(widget-solana-provider): update solana provider to match latest interface

tomiiide:
https://lifi.atlassian.net/browse/LF-13181
Related to https://github.com/lifinance/sdk/pull/327

Since `SDKSolanaProvider` only deals with `Transaction` types from  `@solana/kit`, the widget still uses web3.js, we have convert the transaction types.

### DIFF
--- a/packages/widget-provider-solana/src/providers/SolanaProviderValues.tsx
+++ b/packages/widget-provider-solana/src/providers/SolanaProviderValues.tsx
@@ -15,7 +15,7 @@ export const SolanaProviderValues: FC<
   const {
     wallets,
     selectedWallet: currentWallet,
-    select,
+    connect,
     disconnect,
     connected,
   } = useWallet()
@@ -85,15 +85,15 @@ export const SolanaProviderValues: FC<
       onSuccess?: (address: string, chainId: number) => void
     ) => {
       try {
-        await select(walletName)
-        if (accountAddress) {
-          onSuccess?.(accountAddress, ChainId.SOL)
+        const address = await connect(walletName)
+        if (address) {
+          onSuccess?.(address, ChainId.SOL)
         }
       } catch (error) {
         console.error('Failed to connect wallet:', error)
       }
     },
-    [select, accountAddress]
+    [connect]
   )
 
   return (
--- a/packages/widget-provider-solana/src/wallet-standard/createWalletStandardStore.ts
+++ b/packages/widget-provider-solana/src/wallet-standard/createWalletStandardStore.ts
@@ -62,9 +62,9 @@ export const createWalletStandardStore = ({
         return
       }
 
-      const { select, selectedWalletName } = store.getState()
+      const { connect, selectedWalletName } = store.getState()
       if (selectedWalletName) {
-        select(selectedWalletName, {
+        connect(selectedWalletName, {
           preferredAccount: accounts[0]?.address,
           silent: true,
         })
@@ -112,12 +112,12 @@ export const createWalletStandardStore = ({
         accounts: [],
         selectedAccount: null,
 
-        select: async (
+        connect: async (
           walletName,
           { silent = false, preferredAccount } = {}
         ) => {
           if (typeof window === 'undefined') {
-            return
+            return null
           }
 
           const wallet = get().wallets.find((w) => w.name === walletName)
@@ -146,18 +146,22 @@ export const createWalletStandardStore = ({
               (a) => a.address === targetAccount
             )
 
+            const resolvedAccount = targetExists
+              ? targetAccount
+              : (accounts[0]?.address ?? null)
+
             set({
               selectedWallet: wallet.wallet,
               selectedWalletName: wallet.name,
               connected: true,
               connecting: false,
               accounts,
-              selectedAccount: targetExists
-                ? targetAccount
-                : (accounts[0]?.address ?? null),
+              selectedAccount: resolvedAccount,
             })
 
             subscribeToWalletEvents(api)
+
+            return resolvedAccount
           } catch (e) {
             set({
               selectedWallet: null,
@@ -167,7 +171,10 @@ export const createWalletStandardStore = ({
               accounts: [],
               selectedAccount: null,
             })
-            throw e
+            if (!silent) {
+              throw e
+            }
+            return null
           }
         },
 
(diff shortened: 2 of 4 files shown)


packages/widget-provider-solana/src/providers/SolanaProviderValues.tsx
```diff
@@ -84,8 +88,31 @@ export const SolanaProviderValues: FC<
         isEnabled: true,
         account,
         sdkProvider: SolanaSDKProvider({
-          async getWalletAdapter() {
-            return currentWallet?.adapter as SignerWalletAdapter
+          async getWallet() {
```
effie-ms: Could we have this logic in SDK not to install '@solana/compat', '@solana/transactions', '@solana/addresses'?

effie-ms: (pass only adapter, and then the rest would be already handled via step executor)

tomiiide: I was completely trying to completely remove the `@solana/web3.js` dependency from the sdk. but with the modular arch, might makes sense to move this to the `sdk-solana-provider`

tomiiide: I would move the transaction conversion logic to a helper in the sdk, but I think `signTransaction` should keep its signature

effie-ms: Yeah, I think it's better this way, thanks 🙏

chybisov: Interesting, ideally this should look much simpler, like it was before, when this function just returned the wallet adapter. It may make sense to leave it as `getWalletAdapter` as well.

packages/widget-provider/src/types.ts
```diff
@@ -66,7 +65,7 @@ export type EthereumProviderContext = WidgetProviderContext & {
       chainId,
       skipReady,
     }: {
-      client?: Client
+      client?: any
```
tomiiide, line 68: This expects a viem `Client` type that was previously re-exported from the sdk.

This re-export was removed from the sdk [here](https://github.com/lifinance/sdk/commit/079c553c56921ddbed39280d5b0e42470f73957b#r174624020)

Now, there are three solutions:
1. Keep the re-export _(simple and easy but feels dirty)_
2. Add `viem` as a dependency to `widget-provider`  _(cleaner but increases bundle size for just one type)_
3. Re-architect the widget such that contexts and types are defined in their respective ecosystem provider _(Ideal but requires more work)_

Leaning towards 3, but I would need to think deeply about the 2nd order effects.
`SuiContext` should be in `sui-provider` and be imported by `wallet-provider` however, we don't want `wallet-provider` to depend on any of the ecosystem provider.

So how would `wallet-provider` be aware of other ecosystem contexts without defining and import it ?

I have an idea of `wallet-provider` having a `sdkProvidersStore`  that other ecosystem providers can import and then register/unregister. 

@chybisov @effie-ms Let me know your thoughts on this.

packages/widget-provider-solana/src/wallet-standard/wallet-standard-store.ts
chybisov, line 1: Why did we drift from our usual naming convention for this one?

packages/widget-provider-solana/src/providers/SolanaProvider.tsx
```diff
@@ -1,34 +1,17 @@
 import type { WidgetProviderProps } from '@lifi/widget-provider'
-import { ConnectionContext } from '@solana/wallet-adapter-react'
-import { type PropsWithChildren, useContext } from 'react'
-import { SolanaBaseProvider } from './SolanaBaseProvider'
+import type { PropsWithChildren } from 'react'
+import { useSolanaWalletStandard } from '../wallet-standard/useSolanaWalletStandard'
```
chybisov: We should use .js extensions on file imports.

packages/widget-provider-solana/src/wallet-standard/useSolanaWalletStandard.ts
```diff
@@ -0,0 +1,22 @@
...
+    store = createWalletStandardStore(config)
+  }
+  return store
+}
+
+export function useSolanaWalletStandard(
+  config?: WalletStandardConfig
+): SolanaWalletStandardState {
+  return getStore(config)()
+}
```
chybisov: Please check other stores for examples of how we preserve this across the app lifecycle. We shouldn’t store it as a global variable (use useRef).

packages/widget-provider-solana/src/providers/SolanaProviderValues.tsx
```diff
@@ -50,41 +55,45 @@ export const SolanaProviderValues: FC<
...
-        adapter.once('connect', (publicKey: PublicKey) => {
-          onSuccess?.(publicKey?.toString(), ChainId.SOL)
-        })
+      try {
+        await select(walletName)
+        if (accountAddress) {
+          onSuccess?.(accountAddress, ChainId.SOL)
+        }
+      } catch (error) {
+        console.error('Failed to connect wallet:', error)
```
chybisov, line 93: Did a bit more testing and analysis on this PR, found some issues.

There is a race condition here: after `await select(walletName)` completes, `accountAddress` in the closure is stale, it reflects the value before the connection. The `onSuccess` callback receives `null` or the old address instead of the newly connected address.

packages/widget-provider-solana/src/wallet-standard/createWalletStandardStore.ts
```diff
@@ -0,0 +1,273 @@
...
+              return
+            }
+
+            const wallets = getSolanaWallets()
+            const targetWallet = wallets.find(
+              (w) => w.name === persistedWalletName
+            )
+
+            if (targetWallet && !state.connected) {
+              state.select(persistedWalletName, {
```
chybisov: The `select()` call returns a Promise, but it's not awaited, so if a silent connect fails, the promise rejection goes unhandled.

packages/widget-provider-solana/src/providers/SolanaProviderValues.tsx
tomiiide: Good catch.

1. I've renamed the `select` function to `connect` to be more accurate.
2. I've updated the function to return directly the address after connection.

packages/widget-provider-solana/src/wallet-standard/createWalletStandardStore.ts
tomiiide: Good catch here, it's now awaited.

PR #614, opened 2025-12-02
refactor: chains fetching

effie-ms:
https://lifi.atlassian.net/browse/LF-16927

### DIFF
--- a/packages/widget/src/i18n/fr.json
+++ b/packages/widget/src/i18n/fr.json
@@ -127,6 +127,7 @@
       "deleteTransactionHistory": "L'historique d transaction est uniquement stocké localement et ne peut être récupéré si vous le supprimez.",
       "fundsLossPrevention": "",
       "highValueLoss": "La valeur des jetons reçus est nettement inférieure à celle des jetons échangés et au coût de la transaction.",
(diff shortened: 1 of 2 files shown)


packages/widget/src/hooks/useAvailableChains.ts
effie-ms: TODO: `getChains(client)` if external config and `client.getChains()` if internal

packages/widget/src/hooks/useAvailableChains.ts
```diff
@@ -19,9 +20,33 @@ const supportedChainTypes = [
...
+  const internalClient = useSDKClient()
+
+  const externalClient = externalWidgetConfig
+    ? createClient({
+        ...externalWidgetConfig.sdkConfig,
+        apiKey: externalWidgetConfig.apiKey,
+        integrator:
+          externalWidgetConfig.integrator ?? window?.location.hostname,
+      })
+    : undefined
```
chybisov: I think it should be good to put this external client into some state and update it if something changes, to avoid creating it on every render when it is used externally.

effie-ms: Added memoization

PR #619, opened 2025-12-04
fix: misc all networks fixes

effie-ms:
https://lifi.atlassian.net/browse/LF-16955 (1-5, 7)

### DIFF
--- a/packages/widget/src/components/RouteCard/RouteCard.tsx
+++ b/packages/widget/src/components/RouteCard/RouteCard.tsx
@@ -44,7 +44,8 @@ export const RouteCard: React.FC<
       ? { ...route.fromToken, amount: BigInt(route.fromAmount) }
       : { ...route.toToken, amount: BigInt(route.toAmount) }
   const impactToken: TokenAmount | undefined =
-    subvariant !== 'custom'
+    subvariant !== 'custom' &&
+    !hiddenUI?.includes(HiddenUI.RouteCardPriceImpact)
       ? { ...route.fromToken, amount: BigInt(route.fromAmount) }
       : undefined
 
--- a/packages/widget/src/types/widget.ts
+++ b/packages/widget/src/types/widget.ts
@@ -116,6 +116,7 @@ export enum HiddenUI {
   IntegratorStepDetails = 'integratorStepDetails',
   ReverseTokensButton = 'reverseTokensButton',
   RouteTokenDescription = 'routeTokenDescription',
+  RouteCardPriceImpact = 'routeCardPriceImpact',
   ChainSelect = 'chainSelect',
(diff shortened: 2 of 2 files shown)


packages/widget/src/components/AppContainer.tsx
```diff
@@ -114,6 +114,7 @@ const CssBaselineContainer = styled(ScopedCssBaseline, {
         defaultMaxHeight,
     },
     '&:has(.long-list)': {
+      minHeight: getWidgetMaxHeight(theme),
```
effie-ms, line 117: Fix for 5) (height of widget container with fit-content layout)

packages/widget/src/components/Chains/ChainSearchInput.tsx
```diff
@@ -29,7 +29,15 @@ export const ChainSearchInput = ({
...
   if (inExpansion) {
     return (
-      <Box sx={{ pt: 3, pb: 2, px: 2.5, height: searchHeaderHeight }}>
+      <Box
+        sx={{
+          pt: 3,
+          pb: 2,
+          px: 2.5,
+          height: searchHeaderHeight,
+          boxSizing: 'border-box',
```
effie-ms, line 38: Fix for 4) (space between chains search and list items in sidebar)

packages/widget/src/components/SelectTokenButton/SelectTokenButton.tsx
```diff
@@ -77,7 +83,9 @@ export const SelectTokenButton: React.FC<
               isSelected ? (
                 <TokenAvatar token={token} chain={chain} />
               ) : (
-                <AvatarBadgedDefault />
+                <AvatarBadgedDefault
+                  chain={isAllNetworks ? undefined : chain}
+                />
```
chybisov, line 88: In this case, I assume the icon will be displayed even when no chain has been "touched" yet 👀

effie-ms: Yes, we have both `token` and `chain` (`isSelected` is true), then we show both token and chain icons.
If token is not selected and if (all networks is enabled or no chain is set internally) - we show placeholder for the chain icon, otherwise the chain icon. 
We might preselect the chain programmatically (without "touching") - then it will be still shown - is it what you mean?

chybisov: Yes, if all networks are disabled, then the default chains will be kinda preselected, not sure how big of an issue it is for us tho. In most cases, it should work fine, so maybe we shouldn't complicate the logic here just for that edge case.

chybisov (approved):

PR #677, opened 2026-03-23
feat: new send to wallet functionality

effie-ms:
https://linear.app/lifi-linear/issue/EMB-319/replace-send-to-wallet-input-with-inline-send-to-wallet-button

1) The behaviour of `SendToWalletExpandButton` changes: it is not a toggle anymore. The button is visible only when wallet address is not optional and not set and leads to "Send to wallet" page instead of opening "Send to wallet" input card.

2) Adds "Required" tag to the "Send to wallet" input card.

3) `useSendToWalletStore` is removed: the visibility of the "Send to wallet" input card is determined from the availability of wallet address and whether it is required or not.

https://github.com/user-attachments/assets/84f606b7-99b9-441b-885f-c5e4993d0324

## Linear

- https://linear.app/lifi-linear/issue/EMB-319/update-behavior-of-send-to-wallet-expansion


packages/widget/src/components/ReverseTokensButton/ReverseTokensButton.tsx
```diff
@@ -1,14 +1,12 @@
-import ArrowDownward from '@mui/icons-material/ArrowDownward'
 import ArrowForward from '@mui/icons-material/ArrowForward'
+import type { JSX } from 'react'
```
yolodevz, line 2: some components keep React.FC, this one removes it. Is this a convention?

packages/widget/src/stores/form/useFormRef.ts
```diff
@@ -29,15 +27,6 @@ export const useFormRef = (
           (isToAddressObj ? value?.address : value) ||
           formDefaultValues.toAddress
 
-        // sets the send to wallet button state to be open
-        // if there is an address to display
-        if (address) {
-          setSendToWallet(address)
-        }
-
-        // we can assume that the toAddress has been passed as ToAddress object
```
yolodevz, line 38: check if this is important context to keep

yolodevz (approved):

packages/widget/src/components/ReverseTokensButton/ReverseTokensButton.tsx
effie-ms: It's mixed currently, should not be a blocker.

packages/widget/src/stores/form/useFormRef.ts
effie-ms: yes, reverted, let's keep it

chybisov (approved):

PR #684, opened 2026-04-02
test(e2e): add Playwright E2E suite for 17 widget examples

vinzenzLIFI:

[EMB-331 — CI: Playwright E2E suite for widget examples](https://linear.app/lifi-linear/issue/EMB-331)

Adds a full Playwright E2E suite that verifies all 17 active widget example apps still render and behave correctly when packages or example code changes.

### What this adds

**Test suite** (`e2e/`) with a Component Object Model structure:
- `tests/components/` — reusable component objects (WidgetExchange, SettingsView, TokenSelectorView, PlaygroundSidebar)
- `tests/fixtures/base.fixture.ts` — extended test fixture with widget helpers and `waitForTokens()`
- `tests/profiles/` — specs per profile (standard/routed, iframe, nft)
- `tests/playground/` — smoke tests for the widget playground

**4 example profiles**, each with its own spec:

| Profile | Examples | Assertions |
|---|---|---|
| `standard` | vite, connectkit, privy, privy-ethers, rainbowkit, reown, svelte, zustand-widget-config, vue, nextjs, nextjs15, remix, react-router-7 | Widget root, Exchange heading, From/To buttons, Settings flow, token selector end-to-end |
| `routed` | tanstack-router | Same as standard, navigates to `/widget` first |
| `iframe` | vite-iframe, vite-iframe-wagmi | iframe present, widget root inside frame, Exchange heading, Settings button |
| `nft` | nft-checkout | Widget root, Checkout heading, Pay with section |

**Local test scripts** — full build → serve → test → kill cycle:
```bash
pnpm test:example vite          # single example
pnpm test:examples              # all 17 sequentially
```

**CI workflow** (`e2e-examples.yml`) — triggers on every PR to `main`:
- `detect-changes` job computes which examples to test: isolated `examples/<name>/` change → that example only; shared package / lockfile / `e2e/` change → all 17
- One matrix job per affected example (`fail-fast: false`)
- On failure: Playwright report uploaded as artifact + sticky PR comment listing which examples failed (via `marocchino/sticky-pull-request-comment`)

**Rolldown rc.17 fix** — `vite@8.0.10` introduced a Rolldown regression that causes a circular chunk dependency at runtime (`Class extends value undefined`) in any example using `@wagmi/connectors`. Added `build.rolldownOptions.output.strictExecutionOrder: true` to 8 affected examples. Same fix already applied to `widget-embedded` / `widget-playground-vite` in #721. Will be removed once the upstream fix lands in Rolldown.

### Known broken examples (not tested)

| Example | Issue | Ticket |
|---|---|---|
| `dynamic` | `vite-plugin-env-compatible` doesn't shim `process` globally — Dynamic SDK crashes at runtime | [EMB-349](https://linear.app/lifi-linear/issue/EMB-349) |
| `nuxt` | veaury's React bridge fails in Nuxt SSR production build (`R is not a function`) | [EMB-350](https://linear.app/lifi-linear/issue/EMB-350) |
| `deposit-flow` | Widget's own error boundary fires at runtime — root never mounts | [EMB-351](https://linear.app/lifi-linear/issue/EMB-351) |

Flip `status: 'broken' → 'active'` in `e2e/examples.config.ts` to include them once fixed.

### Selector strategy

Tests use ARIA roles and visible text rather than `data-testid` attributes — the widget is a library and `data-testid` would need to be stripped from production builds (not yet solved). Widget root uses a prefix selector `[id^="widget-app-expanded-container"]` because the full ID includes a React `useId()` suffix that varies per build.

N/A — CI workflow and test suite, no UI changes.

## Linear

- https://linear.app/lifi-linear/issue/EMB-331/create-playwright-smoke-test-suite


e2e/tests/profiles/nft.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+ * "Checkout" rather than "Exchange". "NFT listing not found" is expected when
+ * no NFT params are in the URL — that is normal for this test scenario.
+ */
+
+test.describe('NFT checkout widget smoke', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('/')
+  })
+
+  test('widget root is visible', async ({ page }) => {
```
callMeTheQA, line 18: can all of these tests be moved into test.steps() ?

e2e/tests/fixtures/base.fixture.ts
```diff
@@ -0,0 +1,57 @@
...
+ * ```ts
+ * await Promise.all([
+ *   waitForTokens(page),
+ *   page.goto('/'),
+ * ])
+ * ```
+ */
+export async function waitForTokens(page: Page): Promise<void> {
+  await page.waitForResponse(
+    (res) => res.url().includes('/v1/tokens') && res.status() === 200
```
callMeTheQA, line 24: is it worth here to add timeout? e.g 10-15sec?

`await page.waitForResponse(
  (res) => res.url().includes('/v1/tokens') && res.status() === 200,
  { timeout: 15_000 }
)`

callMeTheQA (approved):

e2e/tests/profiles/nft.spec.ts
vinzenzLIFI: I checked that. we would save execution time, you're right. But I think it is worth to have a separate test for each of the "steps". I will consider grouping into steps, if the anount of tests grows on and we are able to group them into steps semantically.

vinzenzLIFI:
Some tests will fail until https://github.com/lifinance/widget/pull/733 is merged

.github/workflows/e2e-examples.yml
```diff
@@ -0,0 +1,251 @@
...
+            vite connectkit privy privy-ethers rainbowkit reown
+            svelte zustand-widget-config vue nextjs nextjs15
+            remix react-router-7 tanstack-router
+            vite-iframe vite-iframe-wagmi nft-checkout
+          )
+
+          # Example directory mapping (name → examples/<dir>)
+          declare -A EXAMPLE_DIR
+          EXAMPLE_DIR=(
+            [vite]="vite"
```
effie-ms: @vinzenzLIFI I was wondering, is there a way to have the list of all examples in one place?
as I could understand, now if we add one more/remove example, we would need to update multiple lists?

vinzenzLIFI: great idea. consolidated the list of all examples in a single JSON which is parsed in the workflow from now on.

effie-ms (approved):

PR #699, opened 2026-04-13
chore: upgrade MUI from v7 to v9

chybisov:
[EMB-339](https://linear.app/lifi-linear/issue/EMB-339/upgrade-mui-from-v7-to-v9)

MUI v9.0.0 was released (v8 was skipped). This PR bumps `@mui/material`, `@mui/icons-material`, `@mui/system`, and `@mui/material-nextjs` to `^9.0.0` across all packages and examples. `@mui/lab` is pinned to `^9.0.0-beta.2` in `widget-playground` since that's the only v9 release for lab.

The codebase-relevant breaking changes were addressed in-place rather than via codemods because the affected sites were few:

- **Removed legacy `*Outline` icon exports** → renamed `DeleteOutline` → `DeleteOutlined`.
- **System shorthand props removed from `BoxOwnProps` / `TypographyOwnProps` / `StackOwnProps`** — `mb`, `mt`, `px`, `py`, `flex`, `fontSize`, `fontWeight`, `lineHeight`, `display`, `alignItems`, etc. moved into `sx={{ ... }}` at every site flagged by `tsc`.
- **Autocomplete API**: `PopperComponent` → `slots.popper`; widened `isOptionEqualToValue` for `freeSolo` `string | Font` value in `FontsControl`.

While migrating, a latent bug in `SelectChainAndToken` surfaced: the inner `<Box sx={{ display: 'flex' }} {...props}>` was being clobbered by the trailing `{...props}` spread. In v7 the caller passed `mb={2}` as a top-level shorthand prop so nothing collided; once that was migrated into `sx={marginSx}`, the From and To selectors started stacking vertically. Fixed by reordering to `<Box {...props} sx={{ display: 'flex', ...props.sx }}>`.

Verified via A/B chrome comparison of the playground running v7 vs v9 — pixel-equivalent after the `SelectChainAndToken` fix.

## Test plan
- [x] `pnpm check:types` clean across all 12 packages
- [x] `pnpm check:write` (Biome) clean
- [x] `pnpm build` succeeds
- [x] `pnpm --filter @lifi/widget test` — 33/33 passing
- [x] Playground main page renders with From/To side-by-side and bottom CTA + wallet icon inline


packages/widget/src/components/RouteCard/RouteCardEssentials.tsx
```diff
@@ -78,7 +78,7 @@ export const RouteCardEssentials: React.FC<RouteCardEssentialsProps> = ({
               alignItems: 'center',
             }}
           >
-            <IconTypography mr={0.5} fontSize={16}>
+            <IconTypography sx={{ mr: 0.5, fontSize: 16 }}>
               <AccessTimeFilled fontSize="inherit" />
```
yolodevz, line 82: @chybisov and here? Doesn't it need to be wrapped in `sx` too?

```tsx
 <AccessTimeFilled sx={{ fontSize: 'inherit' }} />
```

packages/widget/src/components/TokenList/TokenDetailsSheetContent.tsx
```diff
@@ -75,10 +75,12 @@ export const TokenDetailsSheetContent: React.ForwardRefExoticComponent<
               ) : (
                 <>
                   <Typography
-                    fontWeight={700}
-                    fontSize="24px"
-                    lineHeight="24px"
-                    color="text.primary"
+                    sx={{
+                      fontWeight: 700,
+                      fontSize: '24px',
```
yolodevz, line 80: This isn't using themed values so it's disconnected from the design system. Checking if it's deliberate.

packages/widget/src/components/TransactionDetails.tsx
```diff
@@ -113,7 +113,7 @@ export const TransactionDetails: React.FC<TransactionDetailsProps> = ({
                 cursor: 'pointer',
               }}
             >
-              <IconTypography mr={0.5} fontSize={16}>
+              <IconTypography sx={{ mr: 0.5, fontSize: 16 }}>
                 <LocalGasStationRounded fontSize="inherit" />
```
yolodevz, line 117: `fontSize` inlined instead of wrapped in `sx`

yolodevz (changes requested):
@chybisov 
Left a few questions.

packages/widget/src/components/RouteCard/RouteCardEssentials.tsx
chybisov: They haven't changed that for icons, so it't not part of the migration.

packages/widget/src/components/TokenList/TokenDetailsSheetContent.tsx
chybisov: Why is it disconnected? The text primary should be taken from the theme. Anyway, it's just a migration to a new prop, values haven't changed.

effie-ms (approved):

yolodevz (approved):

PR #701, opened 2026-04-14
feat: animate execution page

yolodevz:


.claude/launch.json
```diff
@@ -0,0 +1,11 @@
+{
+  "version": "0.0.1",
```
effie-ms: Do we need to merge this config for Claude?

packages/widget/src/pages/TransactionPage/StepActionsList.tsx
```diff
@@ -1,10 +1,10 @@
 import type { RouteExtended } from '@lifi/sdk'
```
effie-ms, line 1: Same file as on Transaction details? Could we make it reusable, if so?

knip.json
```diff
@@ -1,7 +1,15 @@
...
+  "ignore": [
+    "examples/**",
+    "packages/widget-embedded/**",
+    "packages/widget-storybook/**",
+    "scripts/**"
+  ],
   "ignoreDependencies": ["@mui/system", "csstype"],
   "workspaces": {
+    "packages/widget": {
+      "ignore": ["src/components/TransactionStatusCard/index.ts"]
```
effie-ms: Could we drop this index file, from this config and in the directory?

package.json
```diff
@@ -26,6 +26,7 @@
     "release:publish:alpha": "pnpm release:publish:build && pnpm -r publish --access public --no-git-checks --tag alpha && pnpm release:clean",
     "release:publish:beta": "pnpm release:publish:build && pnpm -r publish --access public --no-git-checks --tag beta && pnpm release:clean",
     "release:clean": "pnpm -r --parallel --filter './packages/**' --filter '!*-playground-*' --filter '!*-embedded' exec sh -c \"node ../../scripts/postrelease.js && rm -rf *.md\"",
+    "storybook": "pnpm --filter widget-storybook storybook",
```
effie-ms: Do we need to filter it out from the released/published packages as well?

packages/widget/src/components/StepActions/StepActionRow.tsx
```diff
@@ -1,11 +1,108 @@
...
+ *
+ * Single source of truth for "which rows should appear" — consumed by
+ * static step lists (in-page progress and history) and the animated
+ * `TransactionStatusCard`. The returned array is memoized on its inputs.
+ */
+export function useExecutionRows(
+  route: RouteExtended,
+  toAddress?: string
+): ExecutionRow[] {
+  const { getTransactionLink } = useExplorer()
```
effie-ms: Could you check how stable this dependency is? Otherwise there is no use from this memoization :)

vinzenzLIFI:
> 🤖 The following is an extract from the QA Agent Analysis of this PR. Sharing these findings early so they can be addressed before QA execution begins.

---

## Code Inspection Findings

### 🔴 Must address before merge

**`staggerQueue` singleton breaks multi-widget isolation** (`staggerQueue.ts:2-4`)
`let pending`, `let lastFlushAt`, and `let timer` are module-level globals — a single shared queue across all widget instances on a page. When two `LiFiWidget` instances execute routes concurrently, Widget B's `enqueueStaggered` call overwrites Widget A's pending callback. Widget A's route update is silently dropped. Needs either per-instance queue isolation or an explicit decision to drop multi-widget support with documentation.

**`WidgetEvent.RouteExecutionUpdated` is now a breaking change for integrators**
`emitter.emit(RouteExecutionUpdated …)` is now called inside `enqueueStaggered`, so intermediate events fired within the 1200 ms window are silently dropped. Pre-PR: one event per SDK callback. Post-PR: at most one event per 1200 ms, with intermediate ones lost. Integrators tracking per-step progress via this event will silently stop receiving updates — they won't notice until production. Needs a CHANGELOG entry or migration note before merge.

**`.claude/launch.json` accidentally committed**
VS Code debug config was included in the diff. Reviewer already flagged it with no response yet — worth resolving before further review rounds.

---

### 🟡 Should fix

**`getTransactionLink` missing `useCallback` defeats memoization** (`useExplorer.ts:60`, `StepActionRow.tsx`)
`getTransactionLink` is defined inline without `useCallback`, producing a new function reference on every render. `useMemo` in `useExecutionRows` lists it as a dependency, so the memo recomputes on every render. On the transaction page with active timer re-renders this means continuous row re-derivation. Flagged in an open review thread — hasn't been resolved yet.

---

### 🔵 Worth noting

**`visibility: hidden` in `StaggeredRevealTypography` has no timeout fallback** (`StaggeredRevealTypography.tsx:36`)
Text is invisible until `document.fonts.ready` resolves. The promise does resolve on font failure in all modern browsers, so this is safe in practice — but a code comment explaining the assumption would help future readers.

**Index-based `AnimatePresence` keys on route restart** (`ChecklistSection.tsx:59`)
`key={\`row-${index}\`}` is stable for append-only growth but on a route restart, FAILED rows are removed and then re-added at the same indices — producing spurious enter animations. Low visual impact, easy to fix with stable IDs derived from action type or txHash.

---

*Full test plan including 19 test cases, automation guide, and exit criteria available on request.*

.claude/launch.json
chybisov: Ideally we don't need it

packages/widget/src/hooks/useRouteExecution.ts
```diff
@@ -55,59 +62,80 @@ export const useRouteExecution = ({
     state.deleteRoute,
   ])
 
-  const updateRouteHook = (updatedRoute: Route) => {
-    const routeExecution =
-      routeExecutionStoreContext.getState().routes[updatedRoute.id]
-    if (!routeExecution) {
+  const enqueueStaggered = (task: () => void): void => {
+    const q = staggerRef.current
+    q.next = task
```
effie-ms: Could you confirm we are not losing any events when pacing updates on UI?

packages/widget/src/hooks/useRouteExecution.ts
```diff
@@ -55,59 +62,80 @@ export const useRouteExecution = ({
...
+    if (q.scheduled) {
       return
     }
+    q.scheduled = true
+    setTimeout(
```
effie-ms: Cleanup for this timeout would be nice :)

packages/widget/src/hooks/useRouteExecution.ts
```diff
@@ -55,59 +62,80 @@ export const useRouteExecution = ({
...
+      () => {
+        const run = staggerRef.current.next
+        staggerRef.current.next = () => {}
+        staggerRef.current.scheduled = false
+        staggerRef.current.lastAt = Date.now()
+        run()
+      },
+      q.lastAt + STAGGER_INTERVAL - Date.now()
```
effie-ms: With zero `q.lastAt`, it will be negative - maybe Math.max(0, ...) would make it a bit cleaner.

packages/widget/src/hooks/useRouteExecution.ts
```diff
@@ -43,6 +45,11 @@ export const useRouteExecution = ({
   const queryClient = useQueryClient()
   const { account } = useAccount()
   const resumedAfterMount = useRef(false)
+  const staggerRef = useRef({
```
effie-ms: I think it would be nice if we could fully decouple staggering from the logic of receiving execution data from SDK - it's already pretty complicated.

packages/widget/src/hooks/useRouteExecution.ts
```diff
@@ -200,11 +228,39 @@ export const useRouteExecution = ({
...
+      // in the Pending route context, which would leave title undefined and
+      // briefly unmount the text blocks. Clearing error/substatus keeps the
+      // action-row view consistent with the new status.
+      const reset = structuredClone(route)
+      for (const step of reset.steps) {
+        if (step.execution?.status === 'FAILED') {
+          step.execution.status = 'PENDING'
+          for (const action of step.execution.actions ?? []) {
+            if (action.status === 'FAILED') {
+              action.status = 'PENDING'
```
effie-ms: Not sure if we should `updateRoute(reset)` here - let's see if he could omit it when decoupling the animation.

packages/widget-storybook/.storybook/main.ts
```diff
@@ -0,0 +1,12 @@
+import type { StorybookConfig } from '@storybook/react-vite'
```
effie-ms: If we are not going to support a storybook for the repo - I would be nice to drop it.

packages/widget/src/components/ExecutionStatusCard/ExecutionStatusCard.types.ts
```diff
@@ -0,0 +1,17 @@
+import type { ReactNode } from 'react'
```
effie-ms: Could you move this type inline? We don't really follow the convention of creating separate types.tsx files for components. Just for consistency.

packages/widget/src/components/ExecutionStatusCard/index.ts
```diff
@@ -0,0 +1,2 @@
+export { ExecutionStatusCard } from './ExecutionStatusCard.js'
```
effie-ms: The components can be exported right from the files, without this barrel file.

packages/widget/src/pages/TransactionPage/TransactionFailedButtons.tsx
```diff
@@ -10,13 +15,19 @@ interface TransactionFailedButtonsProps {
   deleteRoute: () => void
 }
 
+/** Action buttons shown when route execution fails: retry, delete, and optional contact support. */
```
effie-ms, line 18: Subjective: I think it's nice to have comments around complex logic, trade-offs, etc. Not sure they are necessary when the code is rather self-documenting. Personally, I would clean them up and left only the necessary ones (e.g. for animations, etc.) Otherwise it ends up being too many comments to read (and support) 😅

packages/widget/src/components/ExecutionStatusCard/StatusIcon.tsx
```diff
@@ -0,0 +1,109 @@
...
+ * all exits share the same recession.
+ */
+export function StatusIcon({ route, status }: StatusIconProps): JSX.Element {
+  const iconKey = resolveExecutionIconKey(route, status)
+  const step = route.steps.at(-1)
+
+  return (
+    <Box
+      sx={{
+        position: 'relative',
```
effie-ms: Until we moved away from MUI, these 3+ property styles could be moved to a style file (StatusIcon.style.tsx). Just a suggestion to keep in mind.

packages/widget/src/AppProvider.tsx
```diff
@@ -23,7 +27,11 @@ export const AppProvider: React.FC<PropsWithChildren<WidgetConfigProps>> = ({
               <SDKClientProvider>
                 <WalletProvider providers={config.providers ?? []}>
                   <StoreProvider config={config} formRef={formRef}>
-                    {children}
+                    <LazyMotion features={loadMotionFeatures} strict>
```
effie-ms: Really love the inceptive to reduce the bundle size. ❤️ 
Still wondering if we manage to achieve any worthy optimization with all this code-splitting and lazy loading.

If I understand it correctly, `LazyMotion` allows downloading the features in a separate request. But also `AppProvider` wraps every single page. So we still download the motion features on all the pages, not only on the transaction execution page - it just does it in 2 requests instead of 1. 

2 parallel requests might be better for slower connections in terms of TTI - but I guess we still won't be able to show anything until both chunks are loaded.

So maybe it's not really worthy?
Or should we scope `LazyMotion` just to the execution page (eventually it might end up more "global" again though)?

What do you think? :)

packages/widget/src/hooks/usePacedValue.ts
```diff
@@ -0,0 +1,42 @@
...
+      setValue(live)
+    }
+    const remaining = windowStart.current + DISPLAY_INTERVAL - Date.now()
+    if (remaining <= 0) {
+      show()
+      return
+    }
+    timer.current = setTimeout(show, remaining)
+    return () => clearTimeout(timer.current)
+  }, [live, paced, value])
```
effie-ms: Could we remove `value` from deps not to execute the effect twice?
And use `const valueRef = useRef(live)` instead?

packages/widget/src/motionFeatures.ts
```diff
@@ -0,0 +1,4 @@
+import { domMax } from 'motion/react'
+
+// domMax (not domAnimation) because the execution UI uses layout animations.
+export default domMax
```
effie-ms: Note for the discussion: `domMax` adds a 27KB gzipped chunk (82.70 KB raw / 26.97 KB gzipped) - this is what we get after tree-shaking and bundling.

main bundle is 1,022KB gzipped, so we offload only 27KB to the 2nd request (only 2 percent of the main chunk).

does not sound like a big win, tbh 😅 what do you think? maybe we could just drop it if there is nothing else meaningful I am missing?

effie-ms (changes requested):
Thank you for the cleanup! 
A separate hook for pacing actions really makes it much for maintainable. 

Please check the comments regarding code-splitting and lazy-loading - maybe we could omit them for now, they don't seem to make a significant difference. Otherwise - lgtm.

packages/widget/src/motionFeatures.ts
yolodevz: Okay :)) We can also just move this to the execution page for now and slowly expand as we introduce more animations.

chybisov: Agree with Liza on this. Also, we don't use default exports (isn't it banned by Biome 😄?) and this file seems unnecessary tbh as we can just import `domMax` where we use it.

effie-ms (approved):

pnpm-lock.yaml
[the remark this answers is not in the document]
yolodevz: @AikidoSec pinpoint the place where you see this issue here

packages/widget/src/AppProvider.tsx
chybisov: It would be nice to address this as well, we should not have lazy providers blocking other ones.

PR #721, opened 2026-05-06
fix: add strictExecutionOrder to example apps (rolldown rc.17 regression)

vinzenzLIFI:

n/a — follows up on #718

#718 added `strictExecutionOrder: true` to `widget-embedded` and `widget-playground-vite` to work around a Rolldown rc.17 regression (introduced in `vite@8.0.10`) that causes a circular chunk dependency at runtime:

> `Class extends value undefined is not a constructor or null`

The same fix is needed for all example apps that use `@wagmi/connectors` (walletConnect/reown), as they produce the same circular chunk dependency via `@walletconnect/sign-client → @noble/secp256k1`. Without it, these apps render a blank page.

Affected examples: `vite`, `connectkit`, `privy`, `privy-ethers`, `rainbowkit`, `reown`, `vite-iframe`, `nft-checkout`.

This workaround will be removed once the upstream fix lands in Rolldown.

n/a

### DIFF
--- a/examples/connectkit/vite.config.ts
+++ b/examples/connectkit/vite.config.ts
@@ -4,6 +4,13 @@ import { defineConfig } from 'vite'
 // https://vite.dev/config/
 export default defineConfig({
   plugins: [react()],
+  build: {
+    rolldownOptions: {
+      output: {
+        strictExecutionOrder: true,
+      },
+    },
+  },
   server: {
     port: 3000,
     open: true,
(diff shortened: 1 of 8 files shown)


effie-ms (commented):
Vite has just released a new version with the fixed rolldown: https://github.com/vitejs/vite/issues/22307 - I think it makes sense to update vite instead of this config update. let me do it everywhere (on playground as well) instead in a separate PR. And then this PR can be closed.

effie-ms:
@vinzenzLIFI this should be a better fix for examples: https://github.com/lifinance/widget/pull/722 :)

PR #731, opened 2026-05-15
[EMB-359] feat(ci): add Agent Review Request label trigger for Zeus QA review

vinzenzLIFI:

[EMB-359 — feat: add "Agent Review Request" label and Zeus QA trigger workflow](https://linear.app/lifi-linear/issue/EMB-359/feat-add-agent-review-label-and-zeus-qa-trigger-workflow)

Adds `.github/workflows/qa-review-trigger.yml`, which fires when the **"Agent Review Request"** label is applied to a PR. It swaps the label to **"QA AI Reviewing"** and dispatches the shared `qa-run.yml` workflow in `lifinance/QA` (`scenario=QAReviewEmbeddables.md`, `agent=zeus`), where Zeus runs the review.

This follows the canonical **solver/backend** pattern: source-repo triggers dispatch to the shared `qa-run.yml` rather than a team-specific dispatch workflow. The QA-side counterpart — deleting the old `qa-review-embeddables-dispatch.yml` and moving GitHub PR label management into the Zeus skill — is **QA-101 / lifinance/QA PR #96**, which must land alongside this.

**Implementation notes:**
- **Label swap** (`Agent Review Request` → `QA AI Reviewing`) uses the repo's own `GITHUB_TOKEN`. The QA app token (`QA_APP_ID` / `QA_APP_PRIVATE_KEY`, scoped to `repositories: QA`) is used **only** for the cross-repo dispatch.
- Native `gh api` for the dispatch — no third-party action.
- Restores `Agent Review Request` if the dispatch fails.
- All verdict-dependent label transitions (→ `QA AI Approved`, or staying on `QA AI Reviewing`) are handled by the Zeus skill in the QA repo, not here — a workflow job only sees pass/fail of the process, not the review verdict.

**Prerequisite (org admin):** provision `secrets.QA_APP_ID` and `secrets.QA_APP_PRIVATE_KEY` in `lifinance/widget`, and create the `Agent Review Request` label.

### DIFF
--- a/.github/workflows/qa-review-trigger.yml
+++ b/.github/workflows/qa-review-trigger.yml
@@ -1,0 +1,28 @@
+name: Trigger Zeus QA Review
+
+on:
+  pull_request:
+    types: [labeled]
+
+jobs:
+  trigger-zeus:
+    if: github.event.label.name == 'Agent review'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - name: Create GitHub App token
+        id: app-token
+        uses: actions/create-github-app-token@fee1f7d63c2ff003460e3d139729b119787bc349 # v2
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: lifinance
+
+      - name: Trigger Zeus QA review
+        uses: peter-evans/repository-dispatch@28959ce8df70de7be546dd1250a005dd32156697 # v4.0.1
+        with:
+          token: ${{ steps.app-token.outputs.token }}
+          repository: lifinance/QA
+          event-type: qa-review-embeddables
+          client-payload: '{"pr_url": "${{ github.event.pull_request.html_url }}"}'
--- a/.github/workflows/qa-review-trigger.yml
+++ b/.github/workflows/qa-review-trigger.yml
@@ -16,12 +16,13 @@ jobs:
         env:
           KEY: ${{ secrets.QA_APP_PRIVATE_KEY }}
         run: |
-          echo "pem<<KEYEOF" >> $GITHUB_OUTPUT
-          printf '%s' "$KEY" \
+          FORMATTED=$(printf '%s' "$KEY" \
             | sed 's/-----BEGIN RSA PRIVATE KEY----- /-----BEGIN RSA PRIVATE KEY-----\n/' \
             | sed 's/ -----END RSA PRIVATE KEY-----/\n-----END RSA PRIVATE KEY-----/' \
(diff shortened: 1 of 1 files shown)


vinzenzLIFI:
@lifi-qa-agent — addressing the open medium/high items we're intentionally not changing:

**Issue 1 (High) — Secret names diverge from AC-2**
This is intentional. The ticket was written before the final naming decision was made. `QA_APP_ID` and `QA_APP_PRIVATE_KEY` were chosen over the generic `APP_ID` / `APP_PRIVATE_KEY` because these credentials are exclusively used by the lifi-qa-agent — the `QA_` prefix makes that unambiguous to anyone browsing the secrets list. The org admin has been informed of the correct names. No change needed.

**AC-1 — Label not verifiable from code**
The `Agent review` label exists and is working. The fact that this very QA review was triggered proves it end-to-end. No action needed.

**AC-4 — End-to-end not verifiable statically**
Same as above — the label trigger, dispatch, and Zeus review comment are all live evidence that the full flow works. No action needed.

**Test coverage — Integration smoke test missing**
This workflow is not part of the widget product — it calls out to a separate repo (`lifinance/QA`) where the QA agent itself is tested. A smoke test in this repo is out of scope.

.github/workflows/qa-review-trigger.yml
```diff
@@ -0,0 +1,30 @@
+name: Trigger Zeus QA Review
+
+on:
+  pull_request:
+    types: [labeled]
+
```
effie-ms, line 6: Suggestion:
```
  concurrency:
    group: qa-review-${{ github.event.pull_request.number }}
    cancel-in-progress: true
```

We can add it to handle cases if someone removes and re-applies the label quickly, and two dispatch runs fire. So a concurrency group with `cancel-in-progress: true` would deduplicate.

.github/workflows/qa-review-trigger.yml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  trigger-zeus:
+    if: ${{ github.event.label.name == 'Agent review' }}
+    runs-on: ubuntu-latest
```
effie-ms, line 14: `timeout-minutes: 5`
It will be useful if anything gets stuck, default 6 hours timeout might be long.

.github/workflows/qa-review-trigger.yml
vinzenzLIFI: good idea. addressed

.github/workflows/qa-review-trigger.yml
vinzenzLIFI: true, this can happen. Added the timeout! thank you

effie-ms (approved):

PR #733, opened 2026-05-18
fix: add wagmi v3 dedup override to prevent dual-context runtime error

vinzenzLIFI:

N/A — regression fix discovered during E2E test setup.

After the `chore: bump packages (#728)` PR, two copies of wagmi@3 ended up being installed across the workspace:
- `wagmi@3.6.11` — resolved for `@lifi/widget-provider-ethereum`'s `wagmi: ^3.x` peer dep
- `wagmi@3.6.12` — resolved for example apps that explicitly pin `wagmi: ^3.6.12`

When these examples are built by Vite, both copies end up in the bundle. The `WagmiProvider` from wagmi@3.6.11 and `useConfig` from wagmi@3.6.12 use different React context objects, so `useConfig` cannot find the provider and throws `WagmiProviderNotFoundError` at runtime. This caused the widget error boundary to fire immediately on page load.

The fix adds a single line to `pnpm-workspace.yaml`:
```yaml
wagmi: '>=3.6.12'
```
This collapses all wagmi@3 resolutions to a single 3.6.12 copy across the workspace. No workspace package directly depends on wagmi@2.x, so this does not affect other packages.

N/A — the visible symptom before the fix was an immediate "Something went wrong!" error boundary on affected examples (nft-checkout, vite-iframe, vite-iframe-wagmi).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,7 +2,7 @@ packages:
   - 'packages/*'
   - 'examples/*'
 overrides:
-  wagmi: '>=3.6.12'
+  wagmi: '^3.6.12'
   '@react-native-async-storage/async-storage': '>=2.2.0'
   '@reown/appkit': '>=1.8.18'
   '@reown/appkit>@walletconnect/universal-provider': '>=2.23.9'
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -2,6 +2,7 @@ packages:
   - 'packages/*'
   - 'examples/*'
 overrides:
+  wagmi: '>=3.6.12'
   '@react-native-async-storage/async-storage': '>=2.2.0'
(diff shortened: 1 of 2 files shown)


pnpm-workspace.yaml
```diff
@@ -2,6 +2,7 @@ packages:
   - 'packages/*'
   - 'examples/*'
 overrides:
+  wagmi: '>=3.6.12'
```
effie-ms: I think I would vote for `^3.6.12` here, because usually we update major versions of `wagmi` manually (lots of breaking changes). 
But I don't think we would update it blindly in any case even with `>=`, so for this particular use case it is just a note, optional.

effie-ms (approved):

chybisov:
hey, it was resolved in another PR.

PR #734, opened 2026-05-19
[EMB-381] add CI check for duplicate dependency resolutions

vinzenzLIFI:

[EMB-381
](https://linear.app/lifi-linear/issue/EMB-381/add-ci-check-for-duplicate-dependency-resolutions)

Some packages must be single-copy in a bundle — React context identity, module-level singletons (`@lifi/sdk`), cross-copy `instanceof` checks (`viem`). When two versions end up in the same Vite bundle the symptoms range from a hard crash at mount (`WagmiProviderNotFoundError`) to silent runtime failures (`Execution data not found`).

The check reads `pnpm-lock.yaml` directly (fast, no network) and fails with an actionable error — including a ready-to-paste `pnpm-workspace.yaml` override — when a watched package has more than one resolved version within the same major.

Running it against the current lockfile immediately surfaces a real duplicate: `wagmi@3.6.11` and `wagmi@3.6.12` are both resolved within major 3.

_N/A — CI tooling only._

### DIFF
--- a/.github/workflows/check-deduped-deps.yaml
+++ b/.github/workflows/check-deduped-deps.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches:
       - main
+    paths:
+      - 'pnpm-lock.yaml'
+      - 'scripts/check-deduped-deps.js'
 
 jobs:
   check-deduped-deps:
--- a/.github/workflows/check-deduped-deps.yaml
+++ b/.github/workflows/check-deduped-deps.yaml
@@ -1,0 +1,16 @@
+name: Check deduped deps
+
+on:
+  pull_request:
+    branches:
+      - main
+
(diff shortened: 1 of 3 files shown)


.github/workflows/check-deduped-deps.yaml
```diff
@@ -0,0 +1,16 @@
+name: Check deduped deps
+
+on:
+  pull_request:
+    branches:
+      - main
```
effie-ms, line 6: I think we will need to run it only on lock file changes:
```
     paths:
        - 'pnpm-lock.yaml'
```

effie-ms: And possibly on the script changes: 'scripts/check-deduped-deps.js'

PR #745, opened 2026-05-25
refactor(widget): replace page-level scroll with internal list scroll

effie-ms:

https://linear.app/lifi-linear/issue/EMB-392/fix-widget-scroll-strategy-for-full-heightmobile-layouts

This refactor addresses the bug captured in the ticket and unifies behaviour across scrollable lists.

Previously, list pages (chain select, token select, tools, activities) used two different scroll strategies:
- **ActivitiesPage** used `useListHeight` to calculate a fixed container height with `overflow: auto`, giving internal list scrolling.
- **SelectChainPage** and **SelectEnabledToolsPage** used `FullPageContainer` with `StickySearchInput` (position: fixed in full-height mode), delegating scroll to the outer `CssBaselineContainer`.

The page-level scroll approach caused incorrect `padding-top` calculations on the scrollable container in full-height layout (especially with the playground's mock header enabled), because the fixed-position header height compensation didn't account for the search input also being fixed-positioned.

This PR unifies all list pages to use the ActivitiesPage pattern — `useListHeight` for a measured container height with internal `overflow: auto`. This:
- Removes `FullPageContainer`, `StickySearchInput`, `ChainSearchInput`, `SearchList`, and the sticky search positioning logic
- Eliminates the `adjustForStickySearchInput` workaround in `SearchNotFound`
- Removes `useScrollableOverflowHidden` and the `.full-page-container` CSS class hack in `AppContainer`
- Extracts `useSelectChainListHeight` hook for chain list height calculation (expansion vs non-expansion modes)
- Adds `paddingEnd` to `VirtualizedTokenList` for bottom spacing


packages/widget/src/pages/SelectEnabledToolsPage.tsx
```diff
@@ -47,7 +40,7 @@ const SelectAllCheckbox = memo<SelectAllCheckboxProps>(
           indeterminate={anyCheckboxesSelected && !allCheckboxesSelected}
           onChange={onClick}
           disabled={noCheckboxesAvailable}
-          sx={{ mr: -1.5 }}
+          sx={{ mr: -1.5, height: 20 }}
```
effie-ms, line 43: 20px makes sure the header does not change because of the embedded "select all" checkbox. 
without it, the checkbox causes layout jump (and unnecessary scrollbar appears) in the first milliseconds of render

packages/widget/src/pages/ActivitiesPage/ActivitiesPage.tsx
```diff
@@ -80,7 +81,7 @@ export const ActivitiesPage = (): JSX.Element => {
   }
 
   return (
-    <PageContainer disableGutters style={{ minHeight: 544 }}>
```
effie-ms, line 83: Since we show empty state component, skeletons and cards are large enough now - this minHeight is not necessary

chybisov (commented):
Nice cleanup! Unifying every list page on the measured-height + internal-scroll pattern is the right call, and it's great to see `FullPageContainer`, `StickySearchInput`, `useScrollableOverflowHidden`, and the `:has(.full-page-container)` workaround all go away in one shot. Net −133 LOC and the bug's root cause is gone.

A few small things worth a look before merging:

1. **`useSelectChainListHeight` runs `useListHeight` even in expansion mode** — in that branch the result is discarded, but the `ResizeObserver`, `useLayoutEffect`, and `getBoundingClientRect` calls still fire on every resize, and `useListHeight` will `console.warn` if `Header`/`ScrollableContainer` aren't found. Either add an `enabled` flag to `useListHeight`, or inline the two branches in `SelectChainContent` (it's only one call site).

2. **Hard-coded `theme.spacing(6)` in `WidgetView.style.tsx`** — the comment says `// height of the fixed header`, but this is exactly the kind of decoupled constant that produced the original bug. If the mock header ever changes height it'll silently regress. Could we read it from the mock element's ref, or share a `MOCK_HEADER_HEIGHT` constant between the two?

3. **`searchHeaderHeight = '80px'` lives on in `useSelectChainListHeight`** — non-expansion measures the header via `headerRef`, expansion uses `calc(... - 80px)`. Since the JSX paddings (`pt: 3 + pb: 2 + ~40px input`) just happen to land on 80, this could drift. Measuring in expansion too would make both branches consistent.

Smaller nits:
- The `minHeight: 544` removal on `ActivitiesPage` is a real behavior change for short lists — worth calling out.
- `paddingEnd: 8` in `VirtualizedTokenList` is `theme.spacing(1)` — a brief comment or theme lookup would help future readers.
- `scrollToTop` with `[]` deps is fine but a touch ceremonial; could be inlined.

No tests added — given the matrix here (3 layout modes × fixed/non-fixed header × playground mock header), could you confirm in the description which combinations were verified manually?

Overall LGTM with those tweaks — thanks for tackling this 🙌

effie-ms:
@chybisov thanks for the comments!

I have split the code of `useSelectChainListHeight` for expansion and not - for readability, consistency and as an alternative to `enabled` flag. Also made a constant for `MOCK_HEADER_HEIGHT` to reuse it instead `theme.spacing(6)`.
`paddingEnd: 8` is for virtualizer, so that the last element in the scrollable area was not to close to the bottom edge. Numeric format is expected by the virtualizer.

Re: activities page - push one more improvement there - made it look similar to wallets pages from https://github.com/lifinance/widget/pull/746 height-wise. Demo:
https://github.com/user-attachments/assets/8d200608-b238-40c9-bde9-7b972cab26ba

Testing: yes, verified manually

chybisov (approved):

PR #766, opened 2026-06-04
fix: derive layout mode from persisted container instead of ephemeral store

effie-ms:

https://linear.app/lifi-linear/issue/EMB-421/fix-playground-height-option-showing-default-on-reload

The playground sidebar "Height" label always showed "Default (fit content)" after page reload, even when a restricted height was configured.

**Root cause:** `selectedLayoutId` lived in the `editTools` Zustand store but was never persisted. On reload it reset to `'default'`, and the sync effect that derived it from the persisted container only ran when `HeightDetailView` was mounted.

**Fix:** Remove `selectedLayoutId` from the store entirely and derive the layout mode from `config.theme.container` via `getLayoutMode()` everywhere. Additionally, remove the `usePlaygroundLayoutControls` hook — each component now talks to the config store directly:

- `HeightDetailView` owns layout selection (the switch over layout modes)
- `HeightControl` owns its own input state and reset logic
- `useSidebarNavLabels` derives the layout label from container
- `DeveloperControlsDetailView` derives `isFullHeight` from container

**Removed:**
- `layoutControl` / `setSelectedLayoutId` / `useLayoutValues` from editTools store
- `usePlaygroundLayoutControls` hook (entire file)


tomiiide (approved):
Looks good!

PR #768, opened 2026-06-04
fix: enable React Compiler for @lifi/widget and @lifi/wallet-management

smrnjeet222:

N/A - outside contributor

Enables the **React Compiler** (`babel-plugin-react-compiler@^1.0.0`, target `19`) across the widget toolchain. The widget already runs on React 19, this PR turns on the compiler's automatic memoization.

**Wiring**
- Library packages (`@lifi/widget`, `@lifi/wallet-management`) : added `@rolldown/plugin-babel` + `@babel/core` and registered `babel-plugin-react-compiler` in their `tsdown.config.ts` so the compiler runs at *publish build* time. Consumers receive pre-compiled output and don't need to add the plugin themselves.
- App packages (`widget-embedded`, `widget-playground-vite`) : registered the plugin via `@vitejs/plugin-react`'s `babel.plugins` option so the playground / iframe host benefit too.
- `knip.json` : ignore `babel-plugin-react-compiler` (only referenced as a string by Babel, not imported).

Picked the `tsdown` + `@rolldown/plugin-babel` route over swc/oxc because the React Compiler currently ships only as a Babel plugin, running it through Babel during the rolldown build is the supported path and keeps the rest of the pipeline (rolldown/oxc) untouched.

**Codebase implications (no refactor in this PR)**
- The compiler auto-memoizes components and hook return values. Going forward, new code should **skip manual `useMemo` / `useCallback` / `React.memo` / `forwardRef`** unless there's a measured reason, the compiler handles it.
- Existing usages are **intentionally not refactored** in this PR to keep the diff reviewable and bisectable. can clean them up incrementally (or via a dedicated codemod PR) once validated the compiler.
- `forwardRef` is also unnecessary on React 19 (`ref` is a regular prop), same "don't add new ones, clean up later" rule.

N/A - build-time change, no runtime/UI behavior change.

- [x] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the [public-docs](https://github.com/lifinance/public-docs) repository.

### DIFF
--- a/.changeset/enable-react-compiler.md
+++ b/.changeset/enable-react-compiler.md
@@ -1,0 +1,6 @@
+---
+"@lifi/widget": minor
+"@lifi/wallet-management": minor
+---
+
+Enable the React Compiler (`babel-plugin-react-compiler`, target React 19) in the publish build, so consumers ship pre-compiled, automatically memoized components without configuring Babel themselves.
--- a/knip.json
+++ b/knip.json
@@ -5,7 +5,11 @@
     "packages/widget-embedded/**",
     "scripts/**"
   ],
-  "ignoreDependencies": ["@mui/system", "csstype"],
+  "ignoreDependencies": [
+    "@mui/system",
+    "babel-plugin-react-compiler",
+    "csstype"
+  ],
   "workspaces": {
     "packages/widget-light": {
       "ignore": ["src/config/version.ts"],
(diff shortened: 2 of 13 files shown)


chybisov:
Hey, thanks for the contribution! The DX of the Babel compiler plugin slowing down the whole build/release pipeline is not great rn. We've tried it before and would prefer to stay with a rust-based rolldown. They recently shipped a Rust-based React compiler, so we will wait once it's available in rolldown.

PR #769, opened 2026-06-05
fix(playground): preserve layout settings when switching themes

effie-ms:

[EMB-424](https://linear.app/lifi-linear/issue/EMB-424/fix-playground-height-reset-when-switching-themes)

`setConfigTheme` replaced the entire `config.theme` with the new theme preset's clone, discarding the user's layout container settings (`height`, `maxHeight`, `display`).

The fix destructures only the layout-related container keys from the current config before applying the new theme, then merges them back on top of the new theme's container defaults. This preserves the user's height/layout choices while letting visual container properties (`borderRadius`, `filter`, etc.) update to match the selected theme.


vinzenzLIFI:
minor find: for some reason, when i switch from default to jumper theme while having a max height set, there is a small gap at the bottom of the container. but if i switch from something else to jumper, then it looks clean only for the token container, the chain container still has the gap. 

https://github.com/user-attachments/assets/b0f18cb4-714a-45e0-8451-5afac3e828f6

vinzenzLIFI:
should we persist dark mode in local storage? it changes back to light mode after refresh (not in scope of this PR)

effie-ms:
> minor find: for some reason, when i switch from default to jumper theme while having a max height set, there is a small gap at the bottom of the container. but if i switch from something else to jumper, then it looks clean only for the token container, the chain container still has the gap.

What is happening there: the height of those long lists (e.g. tokens) is calculated dynamically (from the available space). And when we switch a theme, it is not recalculated. But the header's height seems to change because of borders.
Technically, this 2px gap is visible between a theme with the border around search input and a theme without the border. So I am thinking: if I recalculate the height on every theme switch (not to overcomplicate the parsing to find only borders) - it might not be optimal either. Restricting the header's height - also not really an option because we allow customizing the weight of borders.

I am thinking about leaving it as-is. This would be the problem only for the playground - because the integrators set the theme (borders) once and (typically) do not customize it - so they would not get affected.

effie-ms:
> should we persist dark mode in local storage? it changes back to light mode after refresh (not in scope of this PR)

Good question. To be honest, I am not sure what behaviour is more expected by users (persistence or not) 😅 
On reload - it persists, on reset or switching a theme - not sure.

vinzenzLIFI:
I agree with not overcomplicating a solution if it makes things worse. 

Regarding the dark and light mode. When I set dark or light mode directly in the playground settings, it's persisted. however, when I set in the widget, it's not and will fall back to system. this is probably intended, but I wanted to mention anyway.

imho: on global reset everything should be reset back to how it was before. same should be the case for the custom theme settings. if we have a default, that must be recovered.

effie-ms:
> Regarding the dark and light mode. When I set dark or light mode directly in the playground settings, it's persisted. however, when I set in the widget, it's not and will fall back to system. this is probably intended, but I wanted to mention anyway.

I checked v3 - we indeed did not persist it when toggling the dark/light from the widget. It did apply on the DOM level, but not to the local storage.
It's possible to enable though - I have made a separate PR: https://github.com/lifinance/widget/pull/778. But I think it would be worth discussing if we want it in general, before merging.

vinzenzLIFI:
sounds good to me. I approve this PR then and we can discuss further improvements in the new one you opened.

vinzenzLIFI (approved):

tomiiide (approved):
Looks goood!

PR #771, opened 2026-06-08
fix(provider-ethereum): remove Merkle RPC from default mainnet seed

tomiiide:
https://linear.app/lifi-linear/issue/EMB-432

A customer (Bidali) reported intermittent failed/slow Ethereum quotes. The widget builds its default Ethereum client with viem's `http()` and no explicit URL, which resolves to viem's bundled mainnet default, `https://eth.merkle.io` — a single shared public endpoint. In the browser these requests intermittently surfaced as CORS errors and only succeeded after retrying another RPC, adding latency to the quote.

Note on the mechanism: tested directly, `eth.merkle.io` returns a valid CORS preflight (`204` with `access-control-allow-origin: *`), so it does **not** reject preflight unconditionally. The browser CORS failures are most consistent with rate-limiting on the shared endpoint — error responses (e.g. `429`) that omit the `Access-Control-Allow-Origin` header surface in the browser as CORS errors. The exact failing response hasn't been captured yet (the report traces to a Slack thread, not a HAR). Either way, the root problem is relying on a single bundled public default, which this PR fixes.

This change replaces the bundled Merkle default on the mainnet seed chain with CORS-enabled public RPCs and wraps the chain's default RPC list in a viem `fallback` transport, so a failing endpoint fails over per-request instead of stalling the quote.

Scope note: the new `client({ chain })` body applies the `fallback` transport to **every** chain — not just the mainnet seed. Once `useSyncWagmiConfig` runs, chains synced from the LI.FI API also flow through it (their `rpcUrls.default.http` comes from `convertExtendedChain`), so they now get per-request failover too instead of a single `http()` transport.

Alternative considered: keeping Merkle and reordering transports — rejected because it leaves an unreliable shared endpoint as a default rather than removing the dependency on it.

N/A — no UI changes.


effie-ms:
@tomiiide how do you think, should we open an issue on viem as well, so that they updated the mainnet chain definition?

tomiiide:
> @tomiiide how do you think, should we open an issue on viem as well, so that they updated the mainnet chain definition?

@effie-ms I don't think so since the endpoint works, it just has a strict rate limit that appears as a CORS issue.

chybisov (changes requested):
We should not have any hardcoded RPCs in the widget package. Those RPCs are served from the BE and changes to remove Merkle one should be made in data-types repository (@lifi/data-types package) so BE picks up those changes and we would no longer receive it in the FE 🙂

effie-ms:
> We should not have any hardcoded RPCs in the widget package. Those RPCs are served from the BE and changes to remove Merkle one should be made in data-types repository (@lifi/data-types package) so BE picks up those changes and we would no longer receive it in the FE 🙂

Yeah, I was proposing this initially as well. Apparently the BE does not have Merkle endpoint specified anywhere on their side, so it is triggered only for this single mainnet default case, for default wagmi config. 

But maybe there is still a way to pass our chain definition instead, in some way. I would also prefer not to store URLs on the widget side, if it is possible.

tomiiide:
> We should not have any hardcoded RPCs in the widget package. Those RPCs are served from the BE and changes to remove Merkle one should be made in data-types repository (@lifi/data-types package) so BE picks up those changes and we would no longer receive it in the FE 🙂

@chybisov 

I do agree we shouldn't keep an RPC list in the widget, but this case can't be fixed on the BE side as @effie-ms  noted.

1. The BE and `@lifi/data-types` never send Merkle rpc to the widget
2. Merkle only comes from viem's built-in `mainnet` definition, used in `getDefaultWagmiConfig` until `useSyncWagmiConfig` swaps in the BE chains.

I traced the Merkle call to a single spot: `useIsContractAddress` → `getCode`, checking a selected mainnet token through the seed client before sync. Loading with a mainnet token preselected reproduces the exact reported CORS error. 
Two things have to line up for this to happen: 
1. sync hasn't finished (slow `getChains`) 
2. something reads mainnet in that gap. No wallet needed.

**Proposed fix  based on the no hardcoded RPCs hard requirement:**
https://github.com/lifinance/widget/pull/771/commits/a665fcc45469cd27d8dbffec31a9f79773f9a496
1. Give the `mainnet` chain config an empty `rpcUrls.default.http: []`, so Merkle is gone.
2. Real RPCs still come from the BE via `useSyncWagmiConfig`; the client keeps the `fallback` wrapper for the synced chains.
2. Tested with sync delayed 15s: no `eth.merkle.io` requests, no CORS errors, page renders fine.

**One caveat from testing:** with the empty rpc on the  `mainnet` config, a pre-sync `getCode` call throws `UrlRequiredError` that's silently swallowed (no request to log, no `QueryCache` handler, and `useIsContractAddress` maps it to `false`). Normally sync finishes in ms so it self-heals, but on a slow load the 3 default retries (~7s) can run out before sync, and the query won't refetch until a focus/remount.

So I'd pair the seed change with gating `useIsContractAddress` until the chain has a usable RPC (or invalidating `getBytecode` on sync), so the read doesn't run pre-sync at all. 

Let me know what you think?

The only other option is to open an issue on viem and make them update the rpc url.

effie-ms (approved):
I think this option is cleaner than the previous one. And both affected hooks `useIsContractAddress` and `useAddressActivity` are guarded against firing with no chains fetched. So if I am not missing anything, it should be good to go.

vinzenzLIFI:
Apart from the ai review, it seems to look fine. However, I was not able to reproduce the merkle calls pre-fix. There is a race condition to this that I could never trigger. No regressions found though.

vinzenzLIFI (approved):

packages/widget/src/hooks/useIsContractAddress.ts
```diff
@@ -27,7 +32,11 @@ export const useIsContractAddress = (
     refetchInterval: 300_000,
     staleTime: 300_000,
     enabled:
-      chainType === ChainType.EVM && !!chainId && !!address && !!getBytecode,
+      chainType === ChainType.EVM &&
+      !!chainId &&
+      !!address &&
+      !!getBytecode &&
+      isChainAvailable,
```
chybisov: @tomiiide do you think we can skip adding this check here, since this hook is only called inside the components that already implicitly handle this case?

chybisov:
> > We should not have any hardcoded RPCs in the widget package. Those RPCs are served from the BE and changes to remove Merkle one should be made in data-types repository (@lifi/data-types package) so BE picks up those changes and we would no longer receive it in the FE 🙂
> 
> @chybisov
> 
> I do agree we shouldn't keep an RPC list in the widget, but this case can't be fixed on the BE side as @effie-ms noted.
> 
> 1. The BE and `@lifi/data-types` never send Merkle rpc to the widget
> 2. Merkle only comes from viem's built-in `mainnet` definition, used in `getDefaultWagmiConfig` until `useSyncWagmiConfig` swaps in the BE chains.
> 
> I traced the Merkle call to a single spot: `useIsContractAddress` → `getCode`, checking a selected mainnet token through the seed client before sync. Loading with a mainnet token preselected reproduces the exact reported CORS error. Two things have to line up for this to happen:
> 
> 1. sync hasn't finished (slow `getChains`)
> 2. something reads mainnet in that gap. No wallet needed.
> 
> **Proposed fix based on the no hardcoded RPCs hard requirement:** [a665fcc](https://github.com/lifinance/widget/commit/a665fcc45469cd27d8dbffec31a9f79773f9a496)
> 
> 1. Give the `mainnet` chain config an empty `rpcUrls.default.http: []`, so Merkle is gone.
> 2. Real RPCs still come from the BE via `useSyncWagmiConfig`; the client keeps the `fallback` wrapper for the synced chains.
> 3. Tested with sync delayed 15s: no `eth.merkle.io` requests, no CORS errors, page renders fine.
> 
> **One caveat from testing:** with the empty rpc on the `mainnet` config, a pre-sync `getCode` call throws `UrlRequiredError` that's silently swallowed (no request to log, no `QueryCache` handler, and `useIsContractAddress` maps it to `false`). Normally sync finishes in ms so it self-heals, but on a slow load the 3 default retries (~7s) can run out before sync, and the query won't refetch until a focus/remount.
> 
> So I'd pair the seed change with gating `useIsContractAddress` until the chain has a usable RPC (or invalidating `getBytecode` on sync), so the read doesn't run pre-sync at all.
> 
> Let me know what you think?
> 
> The only other option is to open an issue on viem and make them update the rpc url.

Ah, I see, in this case this fix looks fine to me. Please see the comment regarding the chain check.

packages/widget/src/hooks/useIsContractAddress.ts
tomiiide: @chybisov  I think it's better to keep it. Some callers don't actually handle this. Two callers get `chainType` from the connected wallet, not from  `getChainById`:

  - `useGasSufficiency`
  - the from-account call in `useToAddressRequirements`

So `enabled` can flip true before the chains list syncs into the wagmi config. And the components that use them (e.g.  `ReviewButton`) mount and run on startup without waiting for the chains list to load. The widget restores the last-used chain instantly while the chains list loads async, so `getBytecode` can fire for a chain wagmi doesn't know yet, which throws `ChainNotConfiguredError` 

One follow-up worth considering (not blocking): `!!getChainById(chainId)` is a proxy for "wagmi has this chain", read from the LI.FI chains query that *seeds* the config rather than the config itself, so there's a small  render-then-effect gap. We could expose an `isChainSupported` from the ethereum provider backed by `wagmiConfig.chains`  (the exact condition `getBytecode` needs) and gate on that instead:

  ```ts
  // EthereumProviderValues.tsx
  const isChainSupported = useCallback(
    (chainId: number) => wagmiConfig.chains.some((chain) => chain.id === chainId),
    [wagmiConfig]
  )
  ```

  ```ts
  // useIsContractAddress.ts
  const { getBytecode, isChainSupported } = useEthereumContext()
  const isChainAvailable = !!chainId && !!isChainSupported?.(chainId)
  ```

  That removes the seeding gap and drops the `useAvailableChains` dependency here.

chybisov: @tomiiide `useGasSufficiency` takes chainId from the route, so by the time we have the route, we would for sure have chains loaded. 🤔 In case of `useToAddressRequirements` I think we can just make a small change here

```
  const {
    isContractAddress: isFromContractAddress,
    contractCode: fromContractCode,
    isLoading: isFromContractLoading,
    isFetched: isFromContractFetched,
  } = useIsContractAddress(account.address, fromChain?.id, account.chainType) // fromChainId to fromChain?.id
  const {
    isContractAddress: isToContractAddress,
    isLoading: isToContractLoading,
    isFetched: isToContractFetched,
  } = useIsContractAddress(toAddress, toChain?.id, toChain?.chainType) // toChainId to toChain?.id
```

This way, it will take the chain ID only from available chains.

chybisov (approved):

packages/widget/src/hooks/useIsContractAddress.ts
tomiiide: Alright, done!

PR #772, opened 2026-06-09
fix(tron): disable auto-connect on load to prevent ledger modal persistence

effie-ms:
https://linear.app/lifi-linear/issue/EMB-354/ledger-tron-login-popup-persists-after-wallet-reconnect 

### 1. Disable auto-connect on load

The Tron `WalletProvider` persists the selected adapter name (e.g. `"Ledger"`) in `localStorage` under `tronAdapterName`. With `autoConnect` enabled by default, every page load triggers `LedgerAdapter.connect()`, which opens a Preact-rendered modal injected directly into `document.body`. If the connection fails (device unplugged, HID prompt denied), the modal DOM nodes leak — `connect()` only calls `closeConnectingModal()` in the happy path, not in the `finally` block.

Setting `disableAutoConnectOnLoad` prevents auto-connect on initial page load while still allowing it after manual wallet selection (`hasManuallySetName` ref guards this in the upstream hook).

### 2. Pin `@ledgerhq/hw-transport` to 6.27.1

Even after fixing auto-connect, connecting a Tron Ledger device throws `"Class constructor Transport cannot be invoked without 'new'"`. Root cause: pnpm resolves two versions of `@ledgerhq/hw-transport` — 6.27.1 (ES5-compiled, used by `@tronweb3/tronwallet-adapter-ledger`) and 6.35.2 (native ES6 classes, pulled in transitively). At runtime, the ES5-compiled `@ledgerhq/hw-transport-webhid@6.27.1` calls `_super.call(this)` on the ES6 native `Transport@6.35.2` class, which fails because ES6 classes cannot be called without `new`.

The fix adds scoped pnpm overrides forcing both `@ledgerhq/hw-transport-webhid` and `@ledgerhq/hw-app-trx` to resolve `@ledgerhq/hw-transport` to `6.27.1`, ensuring a single ES5-compatible version throughout the dependency tree.

https://github.com/user-attachments/assets/ac64452d-c354-473e-a8c2-962aff0db22d


effie-ms:
**1. [High] Fix scope — jumper.exchange**
Acknowledged. The widget-side fix is correct for standalone widget consumers and any integrator using the widget's built-in TronBaseProvider. For jumper.exchange specifically, a companion change is needed in jumperexchange/jumper-exchange's TronProvider.tsx to add disableAutoConnectOnLoad to its WalletProvider. Will open a follow-up PR there.
https://github.com/jumperexchange/jumper-exchange/pull/2930

**2. [Medium] Formal AC added to EMB-354**
Added acceptance criteria to the Linear ticket:
- After a failed Ledger connection (device unplugged / HID denied), refreshing the page must not auto-open the Ledger modal
- After a failed Ledger connect, selecting a different wallet must proceed without interruption from the Tron Ledger modal
- Manual Ledger wallet selection must still trigger the connect flow normally

**3. [Medium] Test gap**
widget-provider-tron has no test infrastructure currently. The disableAutoConnectOnLoad prop is a pass-through to @tronweb3/tronwallet-adapter-react-hooks's WalletProvider — testing it would effectively be testing the upstream library's behavior. Leaving as-is for now.

Additionally, this PR now includes a second fix: pinning @ledgerhq/hw-transport to 6.27.1 via pnpm overrides to resolve a "Class constructor Transport cannot be invoked without 'new'" error caused by an ES5/ES6 version mismatch in the resolved dependency tree. Without this fix, Ledger connection does not work at all.

PR #775, opened 2026-06-11
feat(widget): add support for navigation tabs (internal)

effie-ms:
https://linear.app/lifi-linear/issue/EMB-451/enable-simple-and-advanced-jumper-modes-on-widget

**_For testing:_** uncomment this line from the code https://github.com/lifinance/widget/pull/775/changes#diff-7fb6dbe81ac16ec6a71b3f658593e2711e2a94393e9e9a57014e4546eb3c97acR63

Header tabs are now **config-driven** through a new internal `_navigationTabs` option (not part of the public API). Each tab key maps to a `variant` / `mode` / `modeOptions`, and the **active tab drives the displayed flow** — those values are *derived* from the active tab, never duplicated into a store.

**Implementation:**
- **`NavigationTabsStore`** replaces the old `SplitModeStore` and holds only `tabs` + `activeTab`. Per-tab `variant`/`mode`/`modeOptions` live in a static lookup and are resolved by util (`getTabMode` / `getTabVariant` / `getTabModeOptions` / `getTabSplitMode`); each field falls back to `config` when a tab leaves it unset.
- **`NavigationTabsStoreProvider`** re-provides an overridden `WidgetContext` below the store, so every existing `useWidgetConfig()` consumer (container width, mode, modeOptions, …) reflects the active tab with no consumer changes.
- **`HeaderTabs`** is a shared, key-based presentational tabs bar; **`HeaderNavigationTabs`** wires it to the store and **`useNavigationTabLabel`** resolves a tab key to its i18n label.
- The split **Swap / Bridge** tabs are served by this same unified store (no separate split-mode store), preserving existing `mode: 'split'` behaviour. `useSplitMode` is derived from the active tab.
- `utils/variant.ts` was renamed to `utils/mode.ts`.

**Tab presets:** `default` → Swap & Bridge (wide), `private` (compact, swap split), `refuel` (wide), `swap-advanced` / `bridge-advanced` (wide split), `limit` (compact), plus the implicit `swap` / `bridge` split tabs.

> [!NOTE]
> The `_navigationTabs` option, the `limit` mode and the advanced tab keys are all `@internal` — not part of the public widget API.

Tabs can be enabled from config: 
https://github.com/user-attachments/assets/82547e6a-7e15-4848-85db-d8afe34ab680
https://github.com/user-attachments/assets/956457ce-2d7b-46fb-95d6-6aa5ec188f90


packages/widget/src/components/AppContainer.tsx
```diff
@@ -47,7 +47,7 @@ export const RelativeContainer: React.FC<
     position: 'relative',
     boxSizing: 'content-box',
     width: '100%',
-    minWidth: theme.breakpoints.values.xs,
+    minWidth: variant === 'jumper' ? 440 : theme.breakpoints.values.xs,
```
effie-ms: NB: new tab-based layout for Jumper requires a wider widget.

packages/widget/src/AppLayout.tsx
```diff
@@ -8,21 +9,27 @@ import { Expansion } from './components/Expansion/Expansion.js'
...
-  const { elementId, buildUrl } = useWidgetConfig()
+  const { elementId, buildUrl, variant } = useWidgetConfig()
   const wideVariant = useWideVariant()
 
   return (
     <AppExpandedContainer
       id={createElementId(ElementId.AppExpandedContainer, elementId)}
       data-version={version}
     >
+      {variant === 'jumper' && (
```
effie-ms: TODO: what to show on smaller screens?

packages/widget-playground/src/defaultWidgetConfig.ts
```diff
@@ -60,6 +60,7 @@ export const widgetBaseConfig: WidgetConfig = {
   ],
   variant: 'wide',
   // mode: 'split',
+  // _navigationTabs: ['default', 'private', 'refuel'], // ['swap-advanced', 'bridge-advanced', 'limit']
```
effie-ms, line 63: For testing: Uncomment to get a view with tabs

effie-ms:
Addressed the review items + rebased on latest `main` (React 19 migration):

- **#4 — label switch default case** ✅ Added a `default` arm to `useNavigationTabLabel` so it always returns a label (never `undefined`) if the key union grows.
- **#1 — `utils.ts` unit tests** ✅ Added `stores/navigationTabs/utils.test.ts` covering the core branches of `getNavigationTabKeys`, `getInitialActiveTab`, `getTabSplitMode`, and the per-tab `variant`/`mode`/`modeOptions` derivation (preset vs. config fallback).
- **#2 / #3 — store & label hook tests** ⏭️ Out of scope here. The widget package has no jsdom/RTL test infra — vitest runs node-env only (see existing `createSettingsStore.test.ts`, which tests the store factory directly, not the provider). The provider's recreation-on-signature-change and `WidgetContext` override genuinely need React rendering, so these belong in a separate change that introduces the rendering test setup.
- **#5 — small-screen tab overflow** ✅ Resolved. The `TODO` is gone; narrow-screen tab layout is intended to be configured by the integrator (Jumper) via theming rather than hardcoded in the widget.
- **#6 — internal types exported** ✅ Accepted as a known limitation, already documented in the changeset. The `_` prefix + `@internal` JSDoc are the agreed convention for now.

chybisov (approved):

PR #776, opened 2026-06-11
feat(widget): add redesigned amount input cards

effie-ms:

[EMB-452](https://linear.app/lifi-linear/issue/EMB-452/add-sendreceive-amount-input-cards-for-jumper-modes)

This adds an **opt-in** redesigned Send/Receive card pair (`AmountInputCardPair`) alongside the existing amount-entry UI, gated behind a new config flag:

```ts
defaultUI: { layout: 'default' | 'cards' } // default: 'default'
```

- `'default'` (default) — the classic chain/token selectors plus the single `AmountInput` are unchanged. Nothing is removed; integrators who don't opt in see no difference.
- `'cards'` — `MainPage` renders `AmountInputCardPair` for the standard swap/bridge flow (every non-custom mode, plus custom `deposit`). The custom-contract flow still renders `SelectChainAndToken`.

Both layouts coexist behind the flag, so the redesign can roll out gradually. The new `defaultUI.layout` is the umbrella for current and future card-based controls.

### New layout (`layout: 'cards'`)

- **Send card** (`SendAmountCard`): large editable input with adaptive font sizing, token pill selector, percentage pills, fiat/token input toggle, and wallet balance display. Supports price (fiat) input mode via the input-mode store. Percentage pills (25/50/75/Max) appear only when a wallet is connected; **Max** deducts the recommended gas reserve on native tokens. The pills reveal with a staggered fade-in on card hover/focus (and stay visible on touch devices).
- **Receive card** (`ReceiveAmountCard`): read-only amount from the best route, with a fiat/token toggle and token pill selector. The footer shows the route's price impact (via `getPriceImpact`, identical to `RouteDetails`) with an info tooltip. Skeletons show only on the initial fetch — a refetch keeps the existing value rather than flashing placeholders — and the price-impact skeleton is suppressed when `hiddenUI.routeCardPriceImpact` is set.
- **Token pill** (`TokenPillButton`): selected state is a neutral chip-style pill; the empty "Select" state is derived from the app's primary `Button` so its theme treatment stays in sync. The chain icon fills as soon as a chain is selected (matching the classic input), even before a token is picked.
- **Swap button** (`SwapButton`): vertical swap arrow between the cards that reverses the from/to chain + token (clearing the amount) and runs the `toAddress` auto-populate / reset — behavior identical to the classic reverse-tokens button. Hidden in `refuel` mode, whose destination is fixed to native gas (reversing is meaningless).
- **Routes preview**: the redundant "Receive" header on the route preview card is hidden under `layout: 'cards'` since the Receive card already carries it.
- **Avatar** (`AvatarBadgedDefault`): gains optional `avatarSize` / `badgeSize` props so the token pill can size its chain avatar and badge.

### New components
- `AmountInputCard/` — `SendAmountCard`, `ReceiveAmountCard`, `AmountInputCardPair`, `PercentageChips`, `BalanceDisplay`, `FiatValueToggle`, shared styles
- `SwapButton/` — vertical swap arrow with overlapping-card design
- `TokenPillButton/` — pill-shaped token selector (chip-style when selected, primary-button call to action when empty)

https://github.com/user-attachments/assets/c936836e-ed60-4116-a5a8-63ed46e9aa61


effie-ms:
## Review feedback addressed

Pushed in `23c71d9b`:

- **#2 E2e selectors** — added `data-testid="widget-{from,to}-token-button"` to `TokenPillButton` and `data-testid="widget-swap-tokens-button"` to `SwapButton`; updated `WidgetView.ts` (`fromButton`/`toButton`/`reverseTokensButton`) to match. All affected specs run the standard flow, so the pills are always present.
- **#3 Changeset** — added `.changeset/emb-452-amount-input-cards.md` (`@lifi/widget` minor).
- **#4 SwapButton a11y** — `aria-label={t('button.swap')}` (reused existing key).
- **#5 Fiat toggle a11y** — `aria-label` on the `ToggleButton` in both `ReceiveAmountCard` and `FiatValueToggle` when interactive.
- **#6 Magic z-index** — extracted `1110` to a documented `swapButtonZIndex` constant.
- **#7 `theme.shape`** — changed to `theme.vars.shape.borderRadiusSecondary` in `TokenPillButton.style.tsx`.
- **#8 Dead `inputMode['to']`** — wired `ReceiveAmountCard` to `useInputModeStore` for `'to'`, so the key is now consumed (toggle state also survives navigation).
- **Extra (regression)** — `PercentageChips` now also hides when `disabledUI.fromAmount` is set, so the chips can't set a disabled amount field.

### #1 Unit tests — not added (justification)

The widget package has **no component-test infrastructure**: no `jsdom`/`happy-dom`, no `@testing-library/react`, no vitest config, and zero `.tsx` tests — the existing suite is node-only pure-function/store tests. Adding rendering tests for these eight provider/hook-dependent components would require standing up a whole DOM test stack and provider render harness, which is out of scope for this PR. The interactive paths are covered by the updated e2e selectors above. Component-level unit tests should be a separate, dedicated effort if the team wants to introduce that layer.

effie-ms:
Addressed all four post-approval issues in `73020480`:

1. **🔴 Critical (AC #2) — `MainPage.tsx`**: the `cards` branch now respects the custom-contract guard. Extracted `showAmountEntry = !custom || modeOptions?.custom?.type === 'deposit'`; in `cards` mode a custom non-deposit flow renders `SelectChainAndToken` (not `AmountInputCardPair`), matching the default layout. Verified across all four mode × layout combinations.
2. **🟠 Medium (AC #5) — `AmountInputEndAdornment.tsx`**: added the wallet-connection guard (`useAccount({ chainType })` + `!account?.isConnected`) mirroring `PercentageChips`, so the chips no longer render (and can't set `"0"`) without a connected wallet.
3. **🟢 Low (regression) — `AmountInputEndAdornment.style.ts`**: `theme.shape` → `theme.vars.shape.borderRadiusSecondary`.
4. **🟠 Scope** — EMB-452 AC #1 updated to reflect the opt-in `defaultUI.layout: 'cards'` delivery (option a). The opt-in flag is an intentional product decision.

Re-applying the **Agent Review Request** label to trigger a re-review.

chybisov (approved):

PR #780, opened 2026-06-12
feat(tron): re-enable WalletConnect adapter

tomiiide:
https://linear.app/lifi-linear/issue/EMB-435/investigate-walletconnect-modal-closing-immediately-in-widget-v4

WalletConnect was previously stubbed out in the Tron provider (createTronAdapters() registered no WC adapter) because of a "No  accounts found in session" error: the upstream @tronweb3/tronwallet-adapters WalletConnect adapter uses the same WalletConnect Cloud project id and origin as the EVM provider, so the EVM connector's persisted eip155 namespaces bled into the Tron session proposal and wallets connected as EVM.

  This re-enables it by registering WalletConnectAdapter through a new resolveTronWalletConnectConfig helper
  (packages/widget-provider-tron/src/config/walletConnect.ts) that:
  - Defaults customStoragePrefix to 'tron' — isolates the Tron WalletConnect storage from the EVM connector's, which is the actual fix
  for the session-bleed bug. Integrators can still override it via options.
  - Mirrors the EVM provider's config contract — walletConnect: true uses the shipped defaults (incl. the shared LI.FI project id, so  it works with no integrator setup), a config object is merged over those defaults, and false/undefined disables it.
  -  Defaults network to 'Mainnet' (capitalized) — lowercase values fall through to an invalid tron:<name> CAIP id upstream.
  - Adapter registration stays conditional (...(walletConnect ? [new WalletConnectAdapter(...)] : [])), so nothing is registered when WC
  is disabled.

## Limitation
From my tests, only two wallets work over wallet connect.
- [TokenPocket connects via Tron WalletConnect ](https://github.com/tronweb3/tronwallet-adapter/issues/3#issuecomment-2251616301). 
- Tronlink

Other wallets confirmed in these issues:
https://github.com/tronweb3/tronwallet-adapter/issues/3
https://github.com/tronweb3/tronwallet-adapter/issues/37
https://github.com/tronweb3/tronwallet-adapter/issues/82#issuecomment-3177916700

I confirmed this by patching the wallet connect library to log the connection details

### Connection logs from Token pocket
Screenshot 2026-06-11 at 8 49 02 PM

### Connection logs from metamask over wallet connect
Metamask only sends EVM connections over walletconnect.
Screenshot 2026-06-11 at 8 48 48 PM

Screenshot 2026-06-11 at 8 48 24 PM

### DIFF
--- a/packages/widget-provider-tron/src/config/walletConnect.ts
+++ b/packages/widget-provider-tron/src/config/walletConnect.ts
@@ -10,17 +10,17 @@ export const resolveTronWalletConnectConfig = (
     return undefined
   }
   const config = walletConnect === true ? undefined : walletConnect
+  const network = config?.network ?? 'Mainnet'
   return {
     ...config,
-    // Must be capitalized; lowercase yields an invalid `tron:<name>` id upstream.
-    network: config?.network ?? 'Mainnet',
+    // Lowercase yields an invalid `tron:<name>` CAIP id upstream.
+    network: `${network.charAt(0).toUpperCase()}${network.slice(1)}`,
     options: {
       projectId: defaultProjectId,
-      // Isolate Tron WC storage from the EVM connector's (shared project id +
-      // origin); otherwise its eip155 namespaces leak into the Tron proposal
-      // and wallets connect as EVM ("No accounts found in session").
-      customStoragePrefix: 'tron',
       ...config?.options,
+      // Forced last: a distinct prefix keeps the EVM connector's eip155
+      // namespaces (shared project id) out of the Tron session proposal.
+      customStoragePrefix: 'tron',
     },
   }
 }
--- a/packages/widget-provider-tron/src/config/walletConnect.ts
+++ b/packages/widget-provider-tron/src/config/walletConnect.ts
@@ -1,0 +1,26 @@
+import type { WalletConnectAdapterConfig } from '@tronweb3/tronwallet-adapters'
+
+// Same project id as the EVM provider — lets `walletConnect: true` work with no setup.
+const defaultProjectId = '5432e3507d41270bee46b7b85bbc2ef8'
+
+export const resolveTronWalletConnectConfig = (
+  walletConnect: WalletConnectAdapterConfig | boolean | undefined
+): WalletConnectAdapterConfig | undefined => {
+  if (!walletConnect) {
+    return undefined
+  }
+  const config = walletConnect === true ? undefined : walletConnect
+  return {
+    ...config,
+    // Must be capitalized; lowercase yields an invalid `tron:<name>` id upstream.
+    network: config?.network ?? 'Mainnet',
+    options: {
+      projectId: defaultProjectId,
+      // Isolate Tron WC storage from the EVM connector's (shared project id +
+      // origin); otherwise its eip155 namespaces leak into the Tron proposal
+      // and wallets connect as EVM ("No accounts found in session").
+      customStoragePrefix: 'tron',
+      ...config?.options,
+    },
+  }
+}
--- a/packages/widget-provider-tron/src/config/walletConnect.ts
+++ b/packages/widget-provider-tron/src/config/walletConnect.ts
@@ -1,7 +1,16 @@
+import { ChainNetwork } from '@tronweb3/tronwallet-abstract-adapter'
 import type { WalletConnectAdapterConfig } from '@tronweb3/tronwallet-adapters'
 
-// Same project id as the EVM provider — lets `walletConnect: true` work with no setup.
+// Shared with the EVM provider so `walletConnect: true` works with no setup.
 const defaultProjectId = '5432e3507d41270bee46b7b85bbc2ef8'
+
+// Upstream matches network names case-sensitively; snap to canonical casing, pass chainIds through.
+const normalizeNetwork = (
+  network: WalletConnectAdapterConfig['network']
+): WalletConnectAdapterConfig['network'] =>
+  Object.values(ChainNetwork).find(
+    (name) => name.toLowerCase() === network.toLowerCase()
+  ) ?? network
 
 export const resolveTronWalletConnectConfig = (
   walletConnect: WalletConnectAdapterConfig | boolean | undefined
@@ -10,16 +19,13 @@ export const resolveTronWalletConnectConfig = (
     return undefined
   }
   const config = walletConnect === true ? undefined : walletConnect
-  const network = config?.network ?? 'Mainnet'
   return {
     ...config,
-    // Lowercase yields an invalid `tron:<name>` CAIP id upstream.
-    network: `${network.charAt(0).toUpperCase()}${network.slice(1)}`,
+    network: normalizeNetwork(config?.network ?? ChainNetwork.Mainnet),
     options: {
       projectId: defaultProjectId,
       ...config?.options,
-      // Forced last: a distinct prefix keeps the EVM connector's eip155
-      // namespaces (shared project id) out of the Tron session proposal.
+      // Forced last: isolates Tron's WC session from the EVM connector's (shared project id).
       customStoragePrefix: 'tron',
     },
   }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1974,6 +1974,9 @@ importers:
       typescript:
         specifier: ^6.0.3
         version: 6.0.3
+      vitest:
+        specifier: ^4.1.8
+        version: 4.1.8(@opentelemetry/api@1.9.1)(@types/node@25.9.1)(vite@8.0.16(@types/node@25.9.1)(esbuild@0.28.0)(jiti@2.7.0)(terser@5.48.0)(yaml@2.9.0))
 
 packages:
 
@@ -9855,8 +9858,8 @@ packages:
   des.js@1.1.0:
     resolution: {integrity: sha512-r17GxjhUCjSRy8aiJpr8/UadFIzMzJGexI3Nmz4ADi9LYSFx4gTBp80+NaX/YsXWWLhpZ7v/v/ubEc/bCNfKwg==}
 
-  deslop-js@0.0.21:
-    resolution: {integrity: sha512-1fYusMl4tDaQ/xdHFtLfDe8kFrEeU0Vu0Pfiy2k/Gd4HSqYlQj1Z7iRqldNneRyuCzNMhwNBvhaL07Urbh5VOA==}
+  deslop-js@0.0.14:
+    resolution: {integrity: sha512-FY0rNK+mdTozXwNmcJAt5tD9hVOkXEX5ac+Mg77KJvnnsoAh3qIUW64PMnofutjeYz7g1vENMbOcrBmD1v4FlA==}
 
   destr@2.0.5:
     resolution: {integrity: sha512-ugFTXCtDZunbzasqBxrK93Ik/DRYsO6S/fedkWEMKqt04xZ4csmnmwGDBAb07QWNaGMAmnTIemsYZCksjATwsA==}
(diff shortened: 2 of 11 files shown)


tomiiide:
### 🔴 C-1 — Out of scope, and the EVM issue is already resolved

  This is built on Thomas Faber's **2026-06-09** hypothesis (the `createWalletConnectElement()` /
  `<w3m-modal>` AppKit collision). That was superseded two days later. On **2026-06-11** the reporting
  partner found and fixed the real cause themselves:

  > *"wagmi v3 (`@wagmi/connectors`) ships the connector backends as optional peer dependencies, so pnpm
  never installed `@walletconnect/ethereum-provider` or `@coinbase/wallet-sdk`. With no backend 
  installed, the connector mounts the modal and tears it down immediately with no error, which matches 
  the flash and close. We added both as direct dependencies and now WalletConnect and Coinbase both 
  work."*

  So the EVM flash-and-close was an **integrator-side missing-dependency issue, not a widget bug**, and
  it's confirmed resolved. The `createWalletConnectElement()` collision was never confirmed as the
  cause. This PR's scope is the *separate* request from that same thread — enabling Tron wallet
  connection over WalletConnect. I'm intentionally **not** touching `EthereumListItemButton.tsx` /
  `elements.ts` against an unverified theory with no reproduction; that would be scope creep. Happy to
  revisit the modal-mounting helper as a standalone cleanup if a real repro surfaces.

  ### 🔴 H-1 — Fixed
  Good catch on the spread order. `customStoragePrefix: 'tron'` is now applied **after**
  `...config?.options`, so it can no longer be overridden — the changeset's "always forces" guarantee
  now matches the code.

  ### 🔴 H-2 — Fixed
  `network` is now normalized (first letter uppercased) regardless of caller input, so a lowercase
  `'mainnet'` resolves to `'Mainnet'` instead of an invalid `tron:<name>` CAIP id.

  ### 🟠 M-1 — Won't add (consistency)
  None of the `widget-provider-*` packages have a unit-test setup — adding one to `widget-provider-tron`
  alone would make it the lone outlier with its own vitest config and dep. I'd rather introduce test
  infra across the provider packages in a dedicated change than bolt it onto this one. The resolver is
  covered by `check:types` and exercised through the playground in the meantime.

  ### 🟠 M-3 — Done
  Changeset now notes that `createTronAdapters` gained an **optional** `WalletConnectAdapterConfig`
  argument and that the change is backwards-compatible (existing callers keep WC disabled).

  ### 🟠 M-2 — Acknowledged
  Public-docs update for `TronProviderConfig.walletConnect` will follow in the public-docs repo; the
  checklist item stays unchecked until then.

  ### 🟢 L-1 — Leaving as-is
  The default `projectId` intentionally duplicates the EVM provider's so `walletConnect: true` works
  with zero integrator setup. A shared constant would couple `widget-provider-tron` to
  `widget-provider-ethereum` for a single string — not worth it. The comment documents the intent.

pnpm-lock.yaml
effie-ms, line 1: Could you revert these lockfile changes if they are unrelated?

packages/widget-provider-tron/src/config/walletConnect.ts
```diff
@@ -0,0 +1,26 @@
...
+): WalletConnectAdapterConfig | undefined => {
+  if (!walletConnect) {
+    return undefined
+  }
+  const config = walletConnect === true ? undefined : walletConnect
+  const network = config?.network ?? 'Mainnet'
+  return {
+    ...config,
+    // Lowercase yields an invalid `tron:<name>` CAIP id upstream.
+    network: `${network.charAt(0).toUpperCase()}${network.slice(1)}`,
```
effie-ms: How do you think, is there a way we could prevent users from passing a network name with the wrong casing?

I mean without capitalizing the first letter on our side. Maybe with some constant/enum? 

I am thinking there is still a possibility they would pass 'MAINNET' - and then it would be invalid as well.

tomiiide: Great catch, addressed in [d395ad0](https://github.com/lifinance/widget/pull/780/commits/d395ad01b6638da6d42f2e1a903696d00cc688a7).

vinzenzLIFI:
**e2e tested the flow with wallet connect on both evm and tron and it works!**

There are a couple of findings I had and I don't know if you can do anything about it @tomiiide. 
Both trace back to the Tron WC adapter, and a lot of it is in the third-party `@tronweb3/walletconnect-tron@4.0.2` wrapper rather than our code:
### Finding 1 — Tron and EVM resolve to the same address
1. Connect any EVM-only wallet (e.g. MetaMask) via WalletConnect.
2. Click "Connect another wallet" → select WalletConnect for Tron.
3. Instead of scanning the QR code using a Tron wallet, scan using the previously used App with EVM wallet again.

#### Result:
Same wallet address is used for EVM and TRON
Screenshot 2026-06-15 at 16 40 25

#### Root cause: 
the wrapper's address extractor doesn't filter by namespace. In `@tronweb3/walletconnect-tron@4.0.2` `lib/esm/adapter.js` (`extractAddressFromSession`, ~L59-71):

```js
const accounts = Object.values(session.namespaces).flatMap(ns => ns.accounts);
const account = accounts[0];                 // first account from ANY namespace
const address = account.split(':')[2];       // "eip155:1:0x95.." -> "0x95.."
```
It takes `accounts[0]` across all namespaces, so an `eip155` account is happily returned as the Tron address. This isn't only cosmetic — it flows into the route request as the Tron sender:

- `useAccount({ chainType })` returns the Tron adapter's account → `account.address` ([useAccount.ts:57](packages/wallet-management/src/hooks/useAccount.ts#L57))
- which becomes `fromAddress` ([useRoutes.ts:116](packages/widget/src/hooks/useRoutes.ts#L116), [useRoutes.ts:228](packages/widget/src/hooks/useRoutes.ts#L228))
So a Tron→X swap can be built with a `0x`… sender. A guard at the Tron provider boundary (reject a connected Tron account whose address isn't a valid base58 `T…`) would stop this regardless of the wrapper. Upstream fix would be to filter to `tron:` accounts in `extractAddressFromSession`.

### Finding 2 — a previously-connected WC wallet disappears after a while; refresh re-opens the QR

**Repro:** connect both an EVM and a Tron wallet via WalletConnect, use the widget for a few minutes. One of the two wallets vanishes from the connected list; on refresh, its QR pops up again.

**Root cause: the two WalletConnect cores silently share one storage namespace.** The branch's "storage isolation" (`customStoragePrefix: 'tron'`) is **ineffective as installed** — the prefix is set but never reaches WalletConnect.

1. We set it in [walletConnect.ts](packages/widget-provider-tron/src/config/walletConnect.ts) (`customStoragePrefix: 'tron'`, ~L23) precisely to isolate Tron from EVM (both use the same project id `5432e3507d41270bee46b7b85bbc2ef8`).
2. But the wrapper drops it — `@tronweb3/walletconnect-tron@4.0.2` `lib/esm/adapter.js:43-48` only forwards 4 fields to `UniversalProvider.init`:
   ```js
   UniversalProvider.init({ projectId, logger: this._options?.logger,
     relayUrl: this._options?.relayUrl, metadata: this._options?.metadata })
   // customStoragePrefix is NOT forwarded
   ```
3. WC keys are namespaced only by that prefix, never by project id. In `@walletconnect/core`:
   ```js
   get storageKey(){ return this.storagePrefix + this.version + this.core.customStoragePrefix + "//" + this.name }
   // customStoragePrefix = opts?.customStoragePrefix ? `:${prefix}` : ""
   ```
   With both cores at `""`, EVM and Tron write to identical keys: `wc@2:client:0.3//session`, `wc@2:core:0.3//keychain`, `//pairing`, `//expirer`. (EVM and Tron pull different core versions, but same key format + same `0.3` store version → full collision.)
4. Each store persists its **entire** in-memory map (`persist()` → `setItem(storageKey, this.values)`), and each core only holds its own session. So whenever either core writes, it overwrites the shared `//session` blob with a version that omits the other ecosystem's session — **last writer wins**.

#### Why refresh → QR

On reload each adapter rehydrates from storage and auto-connects; the Tron adapter does `client.find(getConnectParams(...)).filter(s => s.acknowledged)` in `connect()` (`walletconnect-tron` `adapter.js`, ~L186). Whichever session wasn't the last writer is no longer in storage → `find()` returns `[]` → `appKit.open()` → QR. Which wallet it is depends on persist ordering, matching "one is missing."

#### Why it also disappears live, no refresh (same root cause; exact trigger less pinned)

Once the victim core loses its backing state (symkey/subscription), its still-live session is torn down and it emits `session_delete`. That propagates cleanly: `walletconnect-tron` `adapter.js:141-150` (`session_delete` handler) → `emit('disconnect')` → WC adapter `emit('stateChanged', Disconnect)` → react-hooks `WalletProvider` → `TronProviderValues` → `useAccount` filters it out → wallet vanishes from the menu. It's purely event-driven, no polling in widget code. Plain TTL expiry is **not** the cause (session TTL is 7 days).

#### Notes

- On the **EVM** side specifically, the disconnect tends to surface via `syncWagmiConfig` calling `reconnect(wagmiConfig)` on every chain-list change ([syncWagmiConfig.ts:25](packages/widget-provider-ethereum/src/utils/syncWagmiConfig.ts#L25), [useSyncWagmiConfig.ts:29](packages/widget-provider-ethereum/src/hooks/useSyncWagmiConfig.ts#L29)) — reconnect probes storage, finds the clobbered EVM session gone, and flips to disconnected.
- There's also a `globalThis._walletConnectCore_` singleton (keyed by the same empty prefix) that may merge the two cores' relayer/expirer/pairing — if so, the core-level stores are shared-single rather than clobbered-by-two, but the **session-store** clobber above holds either way. Worth a runtime check.

#### Fix (in our control, doesn't need the third-party wrapper)

Since the Tron wrapper drops the prefix, isolate the **EVM** side instead — set `customStoragePrefix` on the EVM `walletConnect` connector params (wagmi → `@walletconnect/ethereum-provider` *does* forward it to `UniversalProvider.init`). Then EVM writes `wc@2:…:evm//…`, Tron stays `""`, and they no longer collide. Alternatively `patch-package` the Tron wrapper to forward `options.customStoragePrefix` (or upgrade if a fixed version exists).

#### Confirm at runtime

DevTools → Application → **IndexedDB** (`WALLET_CONNECT_V2_INDEXED_DB`) — you'll see only un-prefixed `wc@2:…` keys (no `…:tron//…`), which alone proves the isolation isn't active. Watch `wc@2:client:0.3//session` flip between holding one vs. both sessions as each core persists.

tomiiide:
@vinzenzLIFI  Thanks for the deep root-cause work, both fixed in `e2aa66ee` without touching the wrapper.

**Finding 1.** Added a guard in `TronProviderValues`: the connected address is validated with `SDKProvider.isAddress` (full base58check via `TronWeb`) before it's exposed, and the adapter is disconnected on a non-Tron address so the user can retry. Concrete trigger: MetaMask doesn't support WalletConnect to Tron, it returns the `0x…` Ethereum address regardless of requested network, so picking "connect with Tron" and scanning with MetaMask handed back an EVM address. The widget now rejects it instead of using it as the Tron sender.

**Finding 2.** Forced `customStoragePrefix: 'evm'` on the wagmi `walletConnect` connector (wagmi forwards it down to `CoreTypes.Options`). EVM and Tron now write distinct keys, so the last-writer-wins clobber is gone. Set on the connector, not the default config, since `resolveConfig` replaces rather than merges.

Heads-up for release: existing EVM WalletConnect users get re-prompted once, since the storage key changes. The upstream wrapper bugs are still worth an issue against `@tronweb3/walletconnect-tron`.

effie-ms (approved):

vinzenzLIFI (approved):
Nice, thanks for the fixes. Looks good now

PR #781, opened 2026-06-12
feat(widget): apply per-tab navigation config 

effie-ms:

https://linear.app/lifi-linear/issue/EMB-465/per-tab-config-for-header-navigation-tabs

This PR applies the tab configuration for internally supported navigation tabs for Jumper.

NB: use this SDK type change to pass private flag for "Private" tab: https://github.com/lifinance/sdk/pull/411 

_N/A_


effie-ms:
Thanks for the review. Addressed below:

**1. [Medium] Form fields not cleared on tab switch — not reproducing; already handled.**
Clearing does happen unconditionally on every tab switch, just not via the `reactiveFormValues`/`Object.hasOwn` path that was traced. `HeaderTabs.handleChange` (`packages/widget/src/components/Header/HeaderTabs.tsx:29-33`) calls `setFieldValue('fromAmount', '')`, `setFieldValue('fromToken', '')`, `setFieldValue('toToken', '')` before delegating to `onChange` → `setActiveTab`. This runs for any tab switch regardless of the tab's config (e.g. a tab that only changes `variant`), and `setActiveTab` is only ever reached through this path. So AC 3 is satisfied. No code change.

**2. [Low] Memoize the empty `navigationTabs` reference — declining.**
In this path the `tabConfig` memo only re-returns the already-stable `widgetConfig` reference, so re-execution triggers no re-renders; the sole cost is a `.find()` over an empty array per render. Not worth the added module-level constant.

**3. [Advisory] Document non-effective fields on `NavigationTabConfig.config` — done.**
Added JSDoc noting `sdkConfig`/`theme`/`appearance`/`languages` are read above the tab provider and are not tab-overridable (dfc0e3cb).

PR #790, opened 2026-06-15
fix(tron): disable auto-connect on load to prevent ledger modal persistence

effie-ms:

https://linear.app/lifi-linear/issue/EMB-354/ledger-tron-login-popup-persists-after-wallet-reconnect

### 1. Disable auto-connect on load

The Tron `WalletProvider` persists the selected adapter name (e.g. `"Ledger"`) in `localStorage` under `tronAdapterName`. With `autoConnect` enabled by default, every page load triggers `LedgerAdapter.connect()`, which opens a Preact-rendered modal injected directly into `document.body`. If the connection fails (device unplugged, HID prompt denied), the modal DOM nodes leak — `connect()` only calls `closeConnectingModal()` in the happy path, not in the `finally` block.

Setting `disableAutoConnectOnLoad` prevents auto-connect on initial page load while still allowing it after manual wallet selection (`hasManuallySetName` ref guards this in the upstream hook).

### 2. Pin `@ledgerhq/hw-transport` to 6.27.1

Even after fixing auto-connect, connecting a Tron Ledger device throws `"Class constructor Transport cannot be invoked without 'new'"`. Root cause: pnpm resolves two versions of `@ledgerhq/hw-transport` — 6.27.1 (ES5-compiled, used by `@tronweb3/tronwallet-adapter-ledger`) and 6.35.2 (native ES6 classes, pulled in transitively). At runtime, the ES5-compiled `@ledgerhq/hw-transport-webhid@6.27.1` calls `_super.call(this)` on the ES6 native `Transport@6.35.2` class, which fails because ES6 classes cannot be called without `new`.

The fix adds scoped pnpm overrides forcing both `@ledgerhq/hw-transport-webhid` and `@ledgerhq/hw-app-trx` to resolve `@ledgerhq/hw-transport` to `6.27.1`, ensuring a single ES5-compatible version throughout the dependency tree.


effie-ms:
NB: failed deployment - related to deleting the project from https://github.com/lifinance/widget/pull/772 - the Cloudflare deployment for that PR was failing as well, the reason might be in the constraints for project names - raised with DevOps for further investigation (DO-538).

The newer deployment passes: https://23951d71.widget-fixledgerm.pages.dev/

tomiiide (approved):
Looks good!

effie-ms:
Closing in favor of https://github.com/lifinance/widget/pull/795

PR #791, opened 2026-06-15
fix(widget): respect tool allow-list on first route request

tomiiide:
https://linear.app/lifi-linear/issue/EMB-450/stale-localstorage-allow-list-returns-disabled-exchange-routes

The integrator allow-list (`exchanges.allow` / `bridges.allow`) is enforced only through the `enabledExchanges` / `enabledBridges` store arrays, which are populated asynchronously by the `/tools` fetch. Before that resolves the array is empty (cleared or fresh storage) or stale, and `useRoutes` then either omits the `exchanges`/`bridges` filter entirely (an empty array reads as "no restriction") or sends the stale list. So the first `/advanced/routes` request escapes the allow-list and can return disabled tools (OKX, Kyber, etc.); it self-corrects once `/tools` resolves, which is why only the first request was affected.

Fix:
- Seed `enabled*` from the config allow-list synchronously at store creation (covers empty/cleared localStorage).
- Intersect any persisted state against the config in the persist `merge` (covers a stale persisted list).

This makes the store reflect the allow-list from the first render with no dependency on `/tools` timing. A `useRoutes`-only fallback was considered but rejected: it scatters the config logic into a second place and leaves the persisted store and settings UI inconsistent.

Known limitation: deny-only / no-allow configs cannot be pre-seeded synchronously (the full tool universe is unknown until `/tools`). The reporting integrator uses `allow`, so this is covered.

N/A (request-payload behavior). Repro: with `exchanges: { allow: ['fly'] }`, clear localStorage and block `/tools` in DevTools, then trigger a quote. The first `POST /advanced/routes` body omits `options.exchanges` before the fix and sends `{ allow: ['fly'] }` after.

- [ ] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the public-docs repository. (N/A, no API change)

### DIFF
--- a/.changeset/emb-450-enabled-tools-config-seed.md
+++ b/.changeset/emb-450-enabled-tools-config-seed.md
@@ -1,0 +1,5 @@
+---
+'@lifi/widget': patch
+---
+
+Respect the integrator's `exchanges`/`bridges` allow-list on the first route request. Previously, before the `/tools` fetch populated the enabled-tools store, the initial `/advanced/routes` request omitted the exchange/bridge filter (or sent a stale persisted list), so it could return and render routes from disabled tools. The enabled-tools state is now seeded synchronously from the config allow-list and stale persisted entries are intersected against it on rehydration.
--- a/packages/widget/src/stores/settings/createSettingsStore.test.ts
+++ b/packages/widget/src/stores/settings/createSettingsStore.test.ts
@@ -3,6 +3,8 @@ import type { WidgetConfig } from '../../types/widget.js'
 import { createSettingsStore } from './createSettingsStore.js'
 
 const STORAGE_KEY = 'li.fi-widget-settings'
+// Must match the persist version in createSettingsStore, else merge tests
+// silently become migration tests.
 const STORE_VERSION = 6
 
 const createLocalStorageMock = (): Storage => {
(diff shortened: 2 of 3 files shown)


tomiiide:
Addressed in fe485f74.

**1. Zero-intersection rehydration (Medium)** — Fixed in code, not just tested. The empty-array case is actually unsafe: `useRoutes` treats `enabledExchanges = []` as "no filter" (the `allowExchanges?.length || disabledExchanges.length` guard is falsy), so it omits `options.exchanges` and the request goes out unrestricted, i.e. the original bug. So `merge` now overlays persisted toggles on the config allow-list seed before intersecting; a zero-overlap persisted map keeps the allow-list. New test asserts `enabledExchanges = ['fly']` (not `[]`) for `seedPersistedState({ _enabledExchanges: { okx, kyber } })` + `allow: ['fly']`.

**2. Stale bridges merge path (Medium)** — Added a test seeding `_enabledBridges: { across, hop, stargate }` + `bridges: { allow: ['across'] }`, asserting `enabledBridges = ['across']`.

**3. Deny-only limitation (Low)** — Inline comment added next to `initialSettings`: deny-only configs can't be seeded since the tool universe is unknown until `/tools`.

Also from the notes: `afterEach` now restores the original `globalThis.localStorage`, and `STORE_VERSION` carries a comment that it must match the store's persist version.

effie-ms (approved):

vinzenzLIFI (approved):
Tested this by programatically pre-loading the browser with a saved session containing a broad list of allowed exchanges while the app was configured to only allow `fly`. Before the fix the first swap route request sent all pre-loaded exchanges to the API (disabled exchanges could appear in  results. After the fix the first request correctly sends only the configured exchange regardless of what was saved in the browser.

PR #792, opened 2026-06-16
feat(widget): advanced-mode settings panel with route & slippage pages

effie-ms:
https://linear.app/lifi-linear/issue/EMB-467/advanced-mode-inline-settings-panel-on-main-page-dedicated-route

- **Settings panel** on the main page, shown only for the advanced split tabs (`swap-advanced` / `bridge-advanced`) via `advancedNavigationTabKeys`: compact rows showing live values for the active sub-mode, ordered to match the settings page — **Swap:** Max slippage · Exchanges; **Bridge:** Route priority · Max slippage · Bridges · Exchanges. Each row navigates to its setting. Route priority is omitted for swaps, since it only affects cross-chain routing.
- **Dedicated settings sub-pages** for Route priority and Slippage (previously inline accordion cards); Bridges/Exchanges already had their own pages. These pages scale to the available height and scroll internally, like the other list pages (Bookmarks, etc.).
  - **Route priority:** Best Return (`CHEAPEST`) / Fastest (`FASTEST`), default Best Return — matching the options previously on `main`.
  - **Slippage:** Auto / 0.5% / 1% / Custom. Selecting **Custom** expands a collapsible input styled like the playground sidebar height control.

https://github.com/user-attachments/assets/1f663510-8e74-40a7-a99f-7b34a35264de


effie-ms:
1. Playground CI passes now.

2. Good catch — fixed in [1c76df9](https://github.com/lifinance/widget/pull/792/commits/1c76df9de9c99a511cac94beaad808771a2f103c). 

Fix: on entering Custom mode we now seed the input from the store slippage (previous slippage value or empty if it was "auto"):

```
const handleCustom = () => {
  setCustomMode(true)
  const initialValue = slippage || ''
  if (initialValue) {
    setInputValue(initialValue)
    setValue('slippage', initialValue)
  }
}

```

One thing intentionally left out of scope: the "Custom" checkmark still won't persist across reopen if the committed value happens to match a preset (isCustom is derived from the value, and 0.5/1 are presets). That's a UX/design decision rather than a bug, so flagging separately rather than papering over it here.

chybisov (approved):

PR #793, opened 2026-06-16
feat(widget): add limit order mode

effie-ms:
https://linear.app/lifi-linear/issue/EMB-322

Implement limit order mode as a dedicated header tab. 

Extended API endpoints: https://github.com/lifinance/sdk/pull/410 

_TODO: attach a screen recording of the limit price card, presets, invert toggle, and expiry/partial-fill settings._


effie-ms:
Thanks for the review — here's the disposition of each item.

### Fixed (commit `8ba07dce`)

- **Item 2 — `mergeDefaultFormValues` boolean-false corruption:** the merge now preserves any explicit user value via `userValues[key]?.value !== undefined ? … : default`, so a deliberate `partiallyFillable=false` (or `0`) survives a config-driven `setDefaultValues` instead of reverting to the default.
- **Item 5 — `ReviewButton` label in limit mode:** added `case 'limit'` to both `getButtonText()` switches, using a new `button.review` ("Review") key.
- **Item 6 — `useLimitMarketRate` guard:** added a clarifying comment. The code already parses with `Number(priceUSD)`, so a `"0"`/missing price correctly returns `undefined` (and avoids an `Infinity` rate) — behaviour was correct, intent is now documented. No logic change.
- **Item 9 — `isString(str: any)`:** now `(str: unknown): str is string`.
- **Item 10 — `aria-disabled` on the custom chip:** the `onClick` now early-returns when `disabled`, so `aria-disabled` actually blocks activation. (Used a handler guard rather than `tabIndex` — the wrapper is a `<div>`, not a tab stop, and the inner `<input>` already handles native `disabled` for keyboard focus.)

### Accepted with justification

- **Item 3 — `validUntil` anchoring:** `@lifi/sdk@4.1.0` natively supports `validUntil` / `partiallyFillable` (`getRoutes` resolves them on the backend), so no transport workaround is needed. The value re-anchors on each 60s refetch, which keeps it close to signing time — intended behaviour, not a bug.
- **Item 4 — limit-order success summary:** out of design scope — no limit-specific success screen was designed for this iteration. Not implementing here.
- **Item 7 — preset percentages:** `[0, 1, 5, 10]` matches the current design; the ticket's "+2%/+5%" wording is outdated.
- **Item 11 — `useLinkedLimitFields` unit test:** the widget package has no DOM/React test environment (no jsdom / testing-library / `renderHook`) — all existing tests are pure functions. The hook's derivation math (`deriveReceiveAmount` / `deriveLimitPrice`) is already covered by `limitOrder.test.ts`, so the hook-level test is skipped rather than introducing a full harness for it.

### Intentional / handled elsewhere

- **Item 1 — `_navigationTabs` in the playground:** kept enabled on purpose for now to exercise the tab bar during testing; will restore before merge.
- **Item 8 — public docs:** to be handled in the `public-docs` repo.
- **Item 12 — visual showcase:** PR description will be updated with a recording.

effie-ms:
# Widget flows: navigation tabs, modes & per-tab UI

This doc covers the config surface added on the limit-order branch: `_navigationTabs`,
the per-tab config layering, the new `limit` mode, `defaultUI.layout`, and the
`NavigationTabChanged` event. It shows how to wire up both a **standard** (single-flow)
widget and an **advanced** (tabbed) widget.

> ⚠️ `_navigationTabs`, `mode: 'limit'`, and the internal tab keys (`default`,
> `private`, `refuel`, `swap-advanced`, `bridge-advanced`, `limit`) are marked
> `@internal` — not part of the stable public API yet. Only `swap`/`bridge` tab keys
> and the non-`limit` modes are public.

---

## The two ways to pick a flow

There are two levers, and they don't stack — `_navigationTabs` wins when both are set.

| Lever | What it does | Renders a tab bar? |
| --- | --- | --- |
| `mode` (+ `modeOptions`) | Picks a **single** flow for the whole widget | Only `split` mode with `{ defaultTab }` |
| `_navigationTabs` | Renders a **tab bar**; the active tab drives the flow | Always |

### Resolution order (`getNavigationTabKeys`)

1. If `_navigationTabs` is set and non-empty → render exactly those tabs.
2. Else if `mode: 'split'` **and** `modeOptions.split` is an object (not a pinned
   string) → render the implicit `['swap', 'bridge']` tabs.
3. Else → no tab bar; `mode` alone drives a single flow.

---

## Standard flows (no tab bar)

Set `mode` and leave `_navigationTabs` unset.

```ts
// Combined swap + bridge in one form
const config: WidgetConfig = { mode: 'default' }

// Gas / refuel only
const config: WidgetConfig = { mode: 'refuel' }

// Split, pinned to a single side (no tabs — the string pins it)
const config: WidgetConfig = { mode: 'split', modeOptions: { split: 'swap' } }

// Split with a Swap/Bridge tab bar (object form → implicit tabs)
const config: WidgetConfig = {
  mode: 'split',
  modeOptions: { split: { defaultTab: 'swap' } },
}
```

Available `mode` values: `'default' | 'split' | 'custom' | 'refuel' | 'limit'`
(`'limit'` is internal; `'custom'` is driven by `modeOptions.custom.type`).

---

## Advanced flows (explicit tab bar)

Set `_navigationTabs` to an ordered list of tab keys. The **first** entry is the
initially active tab. Each key is a preset that expands into a `variant` + `mode`
+ `modeOptions` + `defaultUI` + `requiredUI` bundle.

```ts
// The "advanced" preset: two split sides + limit orders
const config: WidgetConfig = {
  _navigationTabs: ['swap-advanced', 'bridge-advanced', 'limit'],
}

// A "simple" tabbed setup
const config: WidgetConfig = {
  _navigationTabs: ['default', 'private', 'refuel'],
}
```

### Tab key reference

Each key resolves to the bundle below. `defaultUI: { layout: 'cards' }` is applied to
every configured tab, so the tabbed experience uses the card-based form layout.

| Key | Label | variant | mode | modeOptions | requiredUI | Notes |
| --- | --- | --- | --- | --- | --- | --- |
| `default` | Swap & Bridge | `wide` | `default` | — | — | Combined form |
| `private` | Private | `compact` | `default` | — | `toAddress: true` | Forces a recipient address |
| `refuel` | Gas | `wide` | `refuel` | — | — | Gas top-up |
| `limit` | Limit | `compact` | `limit` | — | — | Limit-order flow |
| `swap-advanced` | Swap | `wide` | `split` | `{ split: 'swap' }` | — | Same-chain focus |
| `bridge-advanced` | Bridge | `wide` | `split` | `{ split: 'bridge' }` | — | Cross-chain focus |
| `swap` | Swap | — | `split` | `{ split: 'swap' }` | — | Public split key |
| `bridge` | Bridge | — | `split` | `{ split: 'bridge' }` | — | Public split key |

Labels come from i18n (`header.*`) and are overridable via the widget's translation
config.

---

## Per-tab config layering

While a tab is active, its bundle is layered **over** the widget-level config, so the
rest of the widget reads tab-driven values transparently
(`NavigationTabsStoreProvider`). Precedence, per field:

- `mode`, `variant`, `modeOptions` → **tab value, falling back to config** (tab replaces).
- `defaultUI`, `requiredUI` → **shallow-merged, tab wins on collisions**
  (`{ ...config.defaultUI, ...tab.defaultUI }`). Fields the tab doesn't set are
  inherited from the config.

Example — a config-level `defaultUI` survives except where the tab overrides it:

```ts
// config.defaultUI = { transactionDetailsExpanded: true }
// active tab (e.g. 'private') sets defaultUI = { layout: 'cards' }
// → effective defaultUI = { transactionDetailsExpanded: true, layout: 'cards' }
```

Tabs with no `defaultUI`/`requiredUI` (the `swap`/`bridge` split keys) leave the
config's values untouched.

---

## `defaultUI.layout`

New `defaultUI` field controlling the main form layout:

```ts
defaultUI?: {
  transactionDetailsExpanded?: boolean
  navigationHeaderTitleNoWrap?: boolean
  layout?: 'default' | 'cards' // @default 'default'
}
```

- `'default'` — classic chain/token selectors + single amount input.
- `'cards'` — redesigned stacked Send/Receive amount cards (`AmountInputCardPair`)
  with inline token pills and a swap button. All configured navigation tabs set this.

You can opt a standard (tabless) widget into the card layout directly:

```ts
const config: WidgetConfig = { mode: 'default', defaultUI: { layout: 'cards' } }
```

---

## Reacting to tab changes

A `NavigationTabChanged` event fires on the first resolved tab and on every switch:

```ts
import { WidgetEvent, useWidgetEvents } from '@lifi/widget'

const widgetEvents = useWidgetEvents()
widgetEvents.on(WidgetEvent.NavigationTabChanged, ({ tab, previousTab }) => {
  // tab: NavigationTabKey (now active)
  // previousTab: NavigationTabKey | undefined (undefined on first render)
})
```

Remember to clean up with `.off(WidgetEvent.NavigationTabChanged, handler)`.

---

## Playground

`packages/widget-playground/src/defaultWidgetConfig.ts` ships the advanced preset
enabled so you can see it immediately:

```ts
_navigationTabs: ['swap-advanced', 'bridge-advanced', 'limit'],
```

Swap the array (or comment it out and set `mode`) to try other flows.

PR #794, opened 2026-06-16
[9] feat(checkout): use any route for wallet payments, EVM-only

tomiiide:
N/A

This branch builds on `checkout/08-activity-and-polish` and adds the wallet-any-route work plus an EVM-only restriction for checkout.

- **Wallet pays via any route.** The wallet funding flow pays directly from the connected wallet, so it keeps the integrator's full route set (no Intent Factory deposit-only override). Deposit flows (transfer/exchange/cash) still force the IF-only allow-list to surface a deposit-address route.
- **Connected wallets as destination recipients.** The destination step offers already-connected wallets matching the destination ecosystem, and seeds the connected wallet as the default recipient in wallet flows.
- **EVM-only for now.** Checkout forces `chains.types` to EVM-only, cascading to chain lists, token lists, route quotes, and the wallet menu (which now honors `chains.types`). The native gas token is hidden from source-token selection in the IF deposit flows, since IF cannot accept it; the wallet flow keeps full token support.

Token-type filtering is handled naturally by the existing chain-type config rather than a parallel token allow-list, reusing `useTokens`/`useChains`/`useAvailableChains`.

_Not included; UI changes are funding-source/token-list filtering. Recommend a manual pass in the playground to confirm non-EVM chains/wallets/tokens are hidden and an EVM to EVM quote resolves._


vinzenzLIFI:
## Checkout manual testing — findings

Did a first manual pass on this branch. Tested so far:

- **Pay from Wallet** — end-to-end on Ethereum mainnet, completed successfully.
- **Transfer Crypto** (deposit-address / `smartDeposits`) — end-to-end on Ethereum mainnet (hit the refund path; see notes).
- **Deposit with Cash** (Transak) — up to the Transak payment screen.

This is a first pass, not exhaustive — **more testing will follow** (see "Not tested yet" below). Findings so far:

### 🐞 Bugs (widget)

**1. Back from the refund screen lands in a "Transaction details" limbo**
After a deposit refunds, pressing Back navigates to the Transaction Details page, which polls `/v1/status?txHash=<deposit-contract-creation-hash>&fromChain=1&toChain=1` → `404` in a loop. Result: a permanent "Processing transaction" screen with misleading "USDC received / Swapped to stETH" steps that never resolve and contradict the refund.
→ The tx hash being polled is the deposit address's contract-creation tx, not the user's transfer. Once the flow is refunding, it shouldn't poll a tx-hash status (or should use the user's send tx). Related backend gap: the status response omits the user's initial send tx + sender address ([thread](https://lifi-protocol.slack.com/archives/C0B01J8RL3Y/p1780339283231909)).

**2. Cancel-from-Transak → first Back press does nothing**
Repro: Deposit with cash → Transak opens → close Transak (returns to amount screen) → click Back → no reaction; a second click works.
Cause: the cancel handler does `navigate({ to: enterAmount, replace: true })` (`CheckoutTransactionStatusPage.tsx:120`), which replaces the status route with a **second** `/enter-amount` entry sitting right on top of the existing one. The first `router.history.go(-1)` (`Header.tsx:71`) pops to an identical `/enter-amount` → no visible change.
→ Fix: pop back with `router.history.go(-1)` instead of replace-navigating to `enterAmount`.

**3. Cash flow doesn't forward amount/currency to Transak**
Selecting "Deposit with cash" clears the amount (`SelectSourcePage.tsx:232`) and opens the Transak session without a `fiatAmount`, so Transak shows its own default **"You pay 250 EUR"** — including a currency that may not match the one selected in the checkout currency step. We should forward a sensible default amount and the selected currency.

### ℹ️ Known issues (not widget bugs — for reference)

- **Refund on simulation failure is expected** (IF retries, then refunds); the stETH-on-mainnet flakiness that triggered it on develop is known instability → [CORE-360](https://linear.app/lifi-linear/issue/CORE-360/if-develop-stuck-refund-3-underlying-bugs-solver-hits-public-rpc), [IF edge-case thread](https://lifi-protocol.slack.com/archives/C0B01J8RL3Y/p1777550192924089).
- **Transfer-crypto unsupported on Arbitrum** (IF is EVM-mainnet-only for now) → [IF integration channel](https://lifi-protocol.slack.com/archives/C0B01J8RL3Y).
- **Cash refund returns funds to the Transak wallet, not the user** — open design question → [thread](https://lifi-protocol.slack.com/archives/C0AJXUKBPQE/p1779439152720469).

### 🧪 Not tested yet

**This branch's changes still to verify (#794):**
- **EVM-only enforcement** — non-EVM chains, tokens, and wallets (wallet menu) are hidden.
- **Native gas token hidden in IF flows** — ETH absent from the transfer/cash source-token list, but available in pay-from-wallet.
- **Connected wallets as destination recipients** — the recipient picker on the destination step.

**General:**
- **Custom destination address** — typed address / ENS validation (only used a connected wallet so far).
- **Pending / resume** — close mid-deposit and remount (`resumePending`).
- **`onError` / `onClose`** integrator callbacks (`onSuccess` seen via pay-from-wallet).

**Blocked on a test account / payment method:**
- **Connect Exchange (Mesh), e2e** — no exchange test account. Also, connecting a self-custody wallet via Mesh failed: MetaMask never popped up and Mesh showed *"Oops it looks like you denied the connection request"* (nothing was actually denied). Likely the **known multi-wallet-extension collision** — the console showed `window.ethereum` conflicts (*"Cannot redefine property: ethereum"*), already acknowledged ([dev-chains thread](https://lifi-protocol.slack.com/archives/C088UJ56M2B/p1776334618518679)). Needs a re-test with only MetaMask enabled before filing.
- **Deposit with wallet address (Mesh)** — unclear if implemented; avoided wasting funds.
- **Deposit with Cash — post-payment callback** — no test card / Apple Pay / GPay, so tested only up to the Transak screen; funds arrive → IF executes → final status still needs a run.


[CORE-360]: https://lifi.atlassian.net/browse/CORE-360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

vinzenzLIFI:
### 🐛 Mesh deposit flow — UI flicker on the failure→refund transition

Tested the Mesh (exchange) handoff e2e with real funds. Final **"Refund complete"** screen is correct, but on the way there the UI briefly shows funds-received → regresses to a **loading/watching spinner** → then the refund screen.

**Root cause (IF, already known — not a new bug):** the deposit-address `/status` regresses `PENDING → INTENT_SIMULATION_FAILURE → NOT_FOUND → REFUND_IN_PROGRESS → DONE/REFUNDED`. The intermediate `NOT_FOUND` is the IF status-regression Tomi already raised with Georgi → [thread](https://lifi-protocol.slack.com/archives/C0B01J8RL3Y/p1780932539861049). A hotfix landed ~Jun 15 but the simulation-failure path **still regresses** (Tomi's Jun 16 follow-up; reconfirmed in my Jun 18 test).

**Widget resilience gap (this is the part we can fix here):** the status hook treats *any* `NOT_FOUND` as "pre-deposit, keep watching" and discards the last real status — [`useCheckoutTransactionStatus.ts#L123`](https://github.com/lifinance/widget/blob/checkout/09-wallet-any-route/packages/widget-checkout/src/hooks/useCheckoutTransactionStatus.ts#L123):

```ts
const resolvedStatus = data && data.status !== 'NOT_FOUND' ? data : undefined
```

So one regressive `NOT_FOUND` collapses `resolvedStatus`/`phase` and the render falls back to the watching screen (`CheckoutTransactionStatusPage.tsx#L253`), un-showing the real state. Suggested fix: **latch the last non-`NOT_FOUND` status and never downgrade on a later `NOT_FOUND`** (deposit-address path); later real updates still overwrite it. Same guard should cover the resume path (`CheckoutTransactionStatusPage.tsx#L126-148`), which currently bails to enter-amount on `NOT_FOUND` — a manual reload during that window would lose the refund screen.

This makes the UX resilient regardless of the IF fix timeline.

vinzenzLIFI:
### ⚠️ UX gap — a recently-started deposit can be abandoned because it isn't tracked yet

Surfaced while testing the Mesh/Transak deposit flows. This is **not a crash or a logic defect** — the code behaves as written. It's a gap in the user journey: there's a path where a real, in-flight deposit gets dropped and the user is sent back to the start, with no way to recover it from the UI.

**The path that breaks (the scenario to protect against):**
1. User starts a deposit (Connect Exchange / Deposit with Cash, or Transfer Crypto) → widget begins watching for the deposit.
2. User leaves to their wallet / exchange to send the funds.
3. Something interrupts them — a technical hiccup, wallet error, app switch, tab reload — and they come back and **reopen / restart** checkout to continue.
4. They pick up the in-progress deposit to keep tracking it.
5. **But the deposit is very recent**, so IF hasn't found/indexed it yet and returns `NOT_FOUND`.
6. The flow reads that single `NOT_FOUND` as "this deposit never happened," **drops the tracked deposit and aborts back to the start** — even though the funds may already be on their way and could land seconds later.

**Why it's a journey gap, not a one-liner:** `NOT_FOUND` is **ambiguous** — it means *both* "nothing was ever sent" *and* "we just haven't found it yet" (provider / IF indexing delay). The flow currently assumes the first meaning the instant it sees `NOT_FOUND` on resume, which is exactly wrong in the window right after a deposit starts. How a not-yet-found deposit *should* be treated (how long to keep waiting, how the user exits, what's reassured vs. terminal) is a **product/UX decision**, so I'm deliberately not prescribing a fix here.

**Direction (intentionally open):** the guiding principle is *don't abandon a deposit that could still arrive.* `NOT_FOUND` should lean toward "keep waiting"; ending the flow should come from the user (clearly, and non-destructively — the deposit stays recoverable) or from a genuine terminal signal, not from a transient "not yet."

**Where it surfaces (for whoever picks it up — not a prescribed fix):** the resume path lands on the status page (`buildResumeNavigation.ts:69-79`), which on a resumed record with no tx hash treats the first `NOT_FOUND` as terminal — clears the persisted record and redirects to amount entry (`CheckoutTransactionStatusPage.tsx:126-148`). Cash/exchange records are written `confirmed-no-hash` with a `depositAddress` and no hash (`usePendingCheckoutWriter.ts:160-172`), so they fall straight into this path.

**Reproduced live (2026-06-22):** completed a Mesh sandbox deposit → record persisted (`confirmed-no-hash`, exchange) → reloaded → reopened → tapped the "Deposit in progress" card → status poll returned `NOT_FOUND` → the pending record was deleted from localStorage and the user bounced to amount entry. Confirmed it deletes immediately on the first `NOT_FOUND`.

Related (separate, IF-side): the `NOT_FOUND`-after-a-real-status regression already raised with Core → [thread](https://lifi-protocol.slack.com/archives/C0B01J8RL3Y/p1780932539861049). The two compound each other but are distinct.

vinzenzLIFI:
### 🐞 Deposit flows swallow server errors (no error surface, no retry)

Across the checkout deposit flows, a failed API call is **never surfaced as an error** — the user is left on a disabled control or an endless spinner. Three cases, all reproduced by mocking the endpoint → 500:

| Failing call | What the user sees | Why |
|---|---|---|
| **`cex/session`** (Mesh – Connect exchange) | Lands on **"Watching for transaction" forever**; no error, Mesh overlay never opens | Mesh sets `error` (not `failure`) and returns before `isOpen` flips (`MeshHost.tsx:161-183`, `setIsOpen` only at 340). Status page renders only `failure` (`CheckoutTransactionStatusPage.tsx:223`) and `useActiveOnRampDeposit` doesn't expose `error` (`OnRampProvider.tsx:58-86`). |
| **`stepTransaction`** (all deposit flows) | Full quote renders, but the **deposit CTA stays disabled** — no message, no retry | `depositAddress` comes from this call → `null` → CTA's `disabled` (`!route \|\| !depositAddress`) stays true (`useCheckoutFlowQuote.ts:56-78`). |
| **`/v1/status`** (deposit poll) | **"Watching for transaction" forever**, re-polls every ~30s indefinitely, no error | Hook reads only `data`, ignores `isError`; `refetchInterval` only stops on `DONE/FAILED/INVALID` — a 500 has no `data.status`, so it never stops (`useCheckoutTransactionStatus.ts`). |

**Generalizations**
- The Mesh case is **not 500-specific**: 400 / 401 / 403 / 424, network error, and 200-with-missing-`linkToken` all hit the same dead-end. An `apiKey` typo (→ 401/403) silently dead-ends — a likely integrator footgun.
- **Transak is the reference for correct behavior**: a session 500 shows *"Could not start Transak session (500). Try again later."* and closing returns to a retryable review screen.
- Minor, same theme: `/advanced/routes` 500 renders a misleading *"No routes available (low liquidity / amount too low / …)"* — a server error misattributed to the token pair.

**Suggested fix (one treatment covers all):** give the deposit flows a shared "something went wrong — try again" surface, and don't strand on the watching/disabled state when the underlying call errored — e.g. map session / stepTransaction / status failures to the existing `failure` error screen with a `retry`.

chybisov (approved):
Approved. Known findings are tracked and deferred to a single fix pass on feat/widget-checkout after the stack merges; Transak onramp session field to be validated in QA before main.

PR #796, opened 2026-06-19
feat: drop React 18 support and migrate to React 19 idioms

chybisov:

[EMB-456](https://linear.app/lifi-linear/issue/EMB-456/drop-react-18-support-migrate-to-react-19)

Formalizes the move off React 18 — the repo already ran on React 19.2.7, so this makes it official and adopts React 19 idioms:

- `react`/`react-dom` peer deps narrowed `>=18` → `>=19` (`widget`, `wallet-management`, `widget-light`).
- `forwardRef` removed from 11 components (ref passed as a prop; `useImperativeHandle` kept).
- `<Context.Provider>` → `<Context>` and `useContext` → `use()` across packages.
- Header `ResizeObserver` → React 19 ref-cleanup callback.
- The 6 `widget-provider-*` packages use React-19-only APIs (`use()`), so they now declare a `react >=19` peer dependency.

**Bumped `minor`** (changeset): v4 is already released, so the React-version floor is treated as a support-matrix change, not a breaking API change.

The Rust/oxc **React Compiler** was evaluated and **deferred** — it's a no-op in the rolldown bundler path at current versions (experimental); the Babel path was intentionally not used.

Also refreshes a few stale facts in `CLAUDE.md` (separate `docs:` commit).

_No visual change._ Verified: `check:types` + Biome across all packages, 40 widget tests, full build, and browser smoke of the default + drawer-`ref` paths.

- [ ] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the [public-docs](https://github.com/lifinance/public-docs) repository. _(N/A — no Widget API change.)_

### DIFF
--- a/.changeset/drop-react-18-require-react-19.md
+++ b/.changeset/drop-react-18-require-react-19.md
@@ -1,0 +1,13 @@
+---
+'@lifi/widget': minor
+'@lifi/wallet-management': minor
+'@lifi/widget-light': minor
+'@lifi/widget-provider': minor
+'@lifi/widget-provider-ethereum': minor
+'@lifi/widget-provider-solana': minor
+'@lifi/widget-provider-bitcoin': minor
+'@lifi/widget-provider-sui': minor
+'@lifi/widget-provider-tron': minor
+---
+
+Drop React 18 support and require React 19+. The `react`/`react-dom` peer dependency range is narrowed from `>=18` to `>=19`, and the components are modernized to React 19 idioms (refs passed as props instead of `forwardRef`, `use()` for context). The `widget-provider-*` packages now use React-19-only APIs and declare a `react: >=19` peer dependency. Integrators must be on React 19 or newer.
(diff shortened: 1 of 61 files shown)


chybisov:
Addressed both QA items in `28635d67`:

**#1 `ExchangeRateBottomSheet.tsx`** — extracted a stable local `close()` now used by both `useImperativeHandle` and the continue/cancel/close handlers, dropping the `(ref as RefObject)` cast; all three handlers are now `useCallback`.

> Note: I did **not** apply the literal suggestion of calling `bottomSheetRef.current?.close()` in the handlers — that path only closes the sheet and would skip `resolverRef.current?.(value)`, leaving the awaited accept/cancel promise unresolved. The local `close(value, bottomSheetClose)` preserves that resolver behavior while removing the unsafe cast.

**#2 `TokenDetailsSheetContent.tsx`** — stopped forwarding the parent `ref` into the content child; it now takes a plain `onClose` callback (`TokenDetailsSheet` passes `() => bottomSheetRef.current?.close()`), removing the stale `typeof ref !== 'function'` guard entirely.

Verified: `check:types`, Biome, and 40 widget tests pass.

effie-ms (approved):
Looks good! 🚀

PR #804, opened 2026-06-24
fix: remove Phantom from Bitcoin wallet connectors

chybisov:
Phantom deprecated its Bitcoin wallet and removed the injected `window.phantom.bitcoin` provider, so it no longer connects for Bitcoin (and was showing as a dead option in the Bitcoin wallet list).

- `widget-provider-bitcoin`: stop registering the `phantom()` connector in `createDefaultBigmiConfig`.
- `widget-provider`: drop the `app.phantom.bitcoin` case from `isWalletInstalled`.

Phantom for Solana/EVM is unaffected. Verified in the playground: Phantom no longer offers Bitcoin, other BTC wallets remain, and Phantom still connects for Solana/EVM.

Refs [EMB-454](https://linear.app/lifi-linear/issue/EMB-454).

### DIFF
--- a/.changeset/remove-phantom-bitcoin.md
+++ b/.changeset/remove-phantom-bitcoin.md
@@ -1,0 +1,6 @@
+---
+'@lifi/widget-provider-bitcoin': patch
+'@lifi/widget-provider': patch
+---
+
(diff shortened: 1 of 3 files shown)


vinzenzLIFI:
@lifi-qa-agent we do not add the linear issue id to the title. this check will be patched soon. ignore this.

vinzenzLIFI (approved):
looks good to me

PR #819, opened 2026-07-08
feat(widget): add tokens.verified allowlist and mark included tokens as verified

chmanie:
Adds a `tokens.verified` config allowlist (`{ chainId, address }[]`) that marks specific
tokens as trusted, suppressing the yellow "unverified token" warning for long-tail tokens
(e.g. Earn/vault tokens only reachable via search) without forcing them into the
featured/popular categories ([JUMEMB-4](https://linear.app/lifi-linear/issue/JUMEMB-4/add-a-token-verified-allowlist-to-override-the-yellow-unverified-token)).
Additionally, `tokens.include` entries are now treated as verified (mirroring the existing
featured/popular behavior) and deduplicated against API results — previously an included
token that also appeared in search results was rendered twice.

https://linear.app/lifi-linear/issue/JUMEMB-4/add-a-token-verified-allowlist-to-override-the-yellow-unverified-token

I opted for both a `verified` list and to show the `included` tokens as `verified: true` to prefer API data if it exists and only add the verified tag if needbe.

- [x] If this PR modifies the Widget API or adds new features that require documentation, I have updated the documentation in the [public-docs](https://github.com/lifinance/public-docs) repository.

### DIFF
--- a/.changeset/add-verified-token-allowlist.md
+++ b/.changeset/add-verified-token-allowlist.md
@@ -1,0 +1,6 @@
+---
+"@lifi/widget": minor
+"@lifi/widget-light": minor
+---
+
+Add `tokens.verified` config allowlist to mark specific tokens as trusted, suppressing the unverified token warning without adding them to the featured or popular categories. Tokens from `tokens.include` are now also marked as verified.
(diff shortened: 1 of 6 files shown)


mmioana (approved):

chybisov (changes requested):
Looks good overall. Please use this convention for PR names. https://www.conventionalcommits.org/en/v1.0.0/

chybisov (approved):

PR #873, opened 2026-09-10
fix(widget): isolate configurable settings defaults

kriss39:

None.

Configurable defaults like slippage and route priority were stored in a module-level object and updated when a widget config was applied.

That makes those defaults shared across widget instances. For example, rendering one widget with a 1% slippage default could leave that value behind for another widget that did not set one.

This change derives the configurable defaults per settings store instead of mutating the shared defaults. Reset also uses the current widget config, so runtime config changes keep the expected reset behavior.

## Testing

Added regression coverage for:

- keeping slippage and route priority defaults isolated between widget instances
- resetting with defaults from the current widget config

Validation passed with:

- widget regression tests
- `pnpm check`
- full workspace build
- `pnpm check:types`
- circular dependency check
- E2E Playground (dev mode)
- E2E Examples (20/20)
- E2E Playground (4/4 Playwright shards)

N/A.

- [x] No public API or documentation changes are required.

### DIFF
--- a/.changeset/isolate-widget-setting-defaults.md
+++ b/.changeset/isolate-widget-setting-defaults.md
@@ -1,0 +1,5 @@
+---
+'@lifi/widget': patch
+---
+
+Keep configurable route defaults isolated between widget instances.
--- a/packages/widget/src/hooks/useSettingMonitor.ts
+++ b/packages/widget/src/hooks/useSettingMonitor.ts
@@ -32,7 +32,7 @@ export const useSettingMonitor = (): {
   ])
   const { tools } = useTools()
   const config = useWidgetConfig()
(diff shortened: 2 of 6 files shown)


kriss39:
Hi @chybisov, thanks for taking a look. I noticed this was closed together with #871 and #872 without a comment — could you share the reason? I'd like to know whether it's a "we don't take external PRs right now" policy, a disagreement with the approach (e.g. you'd rather keep the shared module-level defaults), or something about the changeset/scope, so I can either adjust or stop sending similar fixes. If the underlying bug (settings defaults leaking between widget instances) is still worth fixing, I'm happy to rework this however you prefer. No worries if not.

PR #874, opened 2026-09-11
feat(widget): explain why a quote returned no routes (JUMEMB-98)

chybisov:

[JUMEMB-98](https://linear.app/lifi-linear/issue/JUMEMB-98/expose-failed-or-filtered-out-routes-reason-when-no-routes-are)

When a quote returns no routes the widget showed one generic sentence, so a user could not tell "your amount is too small" from "this pair is not supported" from "try again in a minute". The API already reports why every candidate was dropped and the SDK already surfaces it — `useRoutes` received `unavailableRoutes` and discarded it.

This maps those reasons onto ten buckets, shows the single most useful one as a card, and offers a one-click fix where the widget can apply one.

### Four layers, so each can be reasoned about alone

| Layer | Responsibility |
| --- | --- |
| `useRoutes` `queryFn` | Classifies in place — the ~590 KB `errors` payload never enters the query cache |
| `utils/routeIssues/` | Pure: no React, no i18n. Flatten → match → extract → dedupe → rank → suppress |
| `useRouteIssueCard` | Turns a bucket into copy and, where possible, an action |
| `RouteIssueCard` | Presentational, `memo`, zero hooks |

**Why classify inside `queryFn`:** one measured no-route response carried ~590 KB of `errors`. Caching that per query key across the 60s refetch is not acceptable. The classifier is pure and `queryFn` already holds everything it needs, so only the compact result is cached.

**Query data shape.** `queryFn` now returns `{ routes, issues }` instead of `Route[]`. Direct cache access was limited to two call sites. Keeping the issues inside the cached response makes them expire atomically with the routes — no second lifecycle to keep in sync.

### An unrecognised reason is the default path, not an edge case

Only `GASLESS_FEE_EXCEEDS_INPUT` is contractually stable; everything else is backend prose that can be reworded at any time. `classifyRouteIssues` returns `[]` — and the card falls back to today's byte-identical generic copy — for a new reason string, a new tool-error code, a reworded reason, an all-suppressed payload, a malformed payload, or any throw. The rule table is a lookup, never an exhaustive `switch`, and each entry is classified inside its own `try` so one bad reason cannot discard the rest. Every matcher fragment is pinned by a test, and an unmatched payload warns in development.

### Alternatives considered

- **Classifying in a hook over the cached payload** — rejected; the payload would sit in the cache.
- **A side store keyed by query key** — rejected; two lifecycles to keep in sync.
- **Several stacked cards with an expander** — built, then removed. The reasons are alternatives across tools, not a conjunction of blockers, and a stack read as a list of things the user must all clear.
- **Reusing `useMessageQueue` / `WarningMessages`** — rejected. That queue explains why an *existing* route cannot execute, from client-side conditions, rendered in the form as severity alerts. This explains why *no* route exists, from a backend payload, rendered in the Receive panel.

### Reuse

`Card` (its per-variant hover is already keyed on `onClick`), `priceToTokenAmount`, `formatTokenPrice`, `formatSlippage`, `wrapLongWords`, `t('format.currency')`, and the existing `t(key as any)` idiom for computed keys. A new `useApplyAmount` hook removes the duplicated limit-mode write that `PercentageChips` had inline.

### Things worth a reviewer's attention

- **A figure is only shown when it is trustworthy.** The backend reports amounts in the bridge leg's own token, which it never names. One captured payload held ratios from 1.74× to 10¹³× for a single request. A figure is derived only when the route path proves the leg still holds the user's token (`stringifyPath` joins swaps with `~`, bridges with `-`). Otherwise the card states the bucket with no number.
- **Contradictions are resolved.** Bridges disagree on range, so "too low" and "too high" can both be reported; whichever carries a real figure wins. `NO_POSSIBLE_ROUTE` is emitted per tool, so beside any other reason it is both noise and untrue — it is dropped unless it is the only signal.
- **The receiver card needs a receiver.** It is suppressed unless the user set one and it differs from the sender, since its fix — send to your own address — is otherwise already the state of the form.
- **Behaviour change:** in `mode: 'custom'`, a contract-call quote that 404s with diagnostics now renders the no-routes screen instead of the error state. Without diagnostics it still throws, so the retry affordance is kept.

> Receive panel, ETH → USDC (Polygon), 1 wei:
>
> **No routes available**
> Here's what got in the way, and what might help.
>
> **Amount is too low**
> Routes for this pair need at least $1.20 to work with. Around 0.00049 ETH should get you there.
> `[ Apply suggestion ]`

Verified in the browser across Ethereum, Polygon, Arbitrum, Base and Solana; in all four themes plus dark mode; and with a connected wallet whose balance is below the suggestion — clicking applies the amount and the widget's own insufficient-funds warning takes over.

### Testing

259 tests (110 new), `check:types` clean on `@lifi/widget` and `@lifi/widget-checkout`, `knip:check` clean. Rules are pinned against a captured real API payload; a live sweep over 12 requests across 7 chains and 6 tokens (~1,600 raw reasons) produced **zero unrecognised reasons**.

### Not done

- The `slippageTooTight` card is covered by unit tests but has never been reproduced in a browser — I could not provoke a path-level slippage rejection on demand.
- Copy is English-only; the other 16 locales carry empty placeholders for Crowdin.


chybisov:
Thanks — all eight items are addressed below. Heads: `eed366064` here, `e10f60e05` on jumperexchange/jumper-widget#117 (same changes, ported).

**1. Ranking / bucketing vs the ticket**
- `recipientNotSupported` is back in the ticket's position, right after `destinationAccountNotReady` (`4530444f5`). The demotion predated the custom-receiver filter, which now drops the bucket unless the user chose a receiver — the noise it guarded against. Your contract-receiver example now leads with the receiver card, pinned in `rules.test.ts` ("ranks a refused receiver ahead of liquidity").
- Kept, and sent to the ticket owner for sign-off on JUMEMB-98 with the evidence: `temporary` below `pairNotSupported` (reported on an eco-only quote that said "Try again" on every retry), and the two temporary fragments suppressed. Ticket §2 suppresses `TOOL_NOT_ALLOWED`, and in every capture `Tool … not applied.` is only that code's message — following §3 literally turns 3 of 63 replayed no-route cards into "a provider is busy, try again". Proposal: https://linear.app/lifi-linear/issue/JUMEMB-98/expose-failed-or-filtered-out-routes-reason-when-no-routes-are#comment-7a21c9c6

**2. AC9 — real payloads per bucket**
New captures, each pinned to the tools that refuse so the result is empty:
- `slippageTooTight` — eisen below its 0.01% floor
- `slippageTooLoose` — mayan / mayanMCTP / symbiosis above their 10% cap
- `blockedBySettings` — allbridge with `allowSwitchChain: false`
- `destinationAccountNotReady` — a Stellar receiver with no USDC trustline

Now 8 of 11 buckets render as a card from a real payload, and 10 of 11 have real prose pinned. The gaps are named in a coverage test in `liveCards.test.ts`: `recipientNotSupported` was only ever captured beside a found route (every request that tripped it found another path), `temporary` only beside a higher-ranked reason, and `gaslessNotAvailable` has no reproduction through the public routes endpoint. The replay also caught a wrong assertion: it assumed every slippage fix raises the setting, while a cap is met by lowering it.

**3. `useRoutes.test.tsx`** (new, 9 tests): contract-call 404 with diagnostics → `{ routes: [], issues }` plus `AvailableRoutes([])`; 404 without diagnostics and a non-404 both rethrow; issues cleared on a later non-404 failure and kept on a 404; empty main routes wait for the relayer and show its route alone; relayer `null` still explains and emits `[]`; contract-call success emits its route. Each case was checked by mutating `useRoutes.ts` — every mutation fails exactly the matching test. The fork's version (15 tests) also covers its exact-output quote and the progressive stream.

**4. `useDebouncedWatch.test.tsx`** (5 tests): an immediate write publishes without the delay; the next keystroke is debounced again; an immediate `fromAmount` write leaves a `toAmount` watcher waiting; clearing publishes at once. Mutation-checked.

**5. Rounding** — suggestions no longer use `roundingMode`; amounts and slippage are rounded on the decimal digits. A test stubs `Intl.NumberFormat` without `roundingMode` and fails on the old code (your 1.02 → 1.0404 case came out as `1`, now `1.1`). The slippage cap had the same defect and is fixed with it.

**6. Dump / print** — both moved to one opt-in report, `scripts/route-issues/report.js`, with its own vitest config. The unit suite no longer writes a file or prints.

**7. AC7** — the ETH → SOL capture now pins the full ranked result (`amountTooLow` / `fromTokenValueFloor`, then `liquidity`, `temporary`), so the one rendered card is exact. The 476-entry sample is not in the repo (its raw form is ~590 KB); an AC wording change is in the same ticket proposal.

**8. `useRouteIssueCard.test.tsx` / `RouteNotFoundCard.test.tsx`** (7 tests): the amount write is immediate; limit mode goes through `setSendAmount`; clearing the receiver also clears the bookmark; retry invalidates every routes query and nothing else; the generic copy stays when there are no issues, and exactly one card shows otherwise. Mutation-checked.

**Checked in the browser** (local playground, production API; the fork against the develop API)
- Amount too low → Apply wrote 0.64 USDC and routes showed within a second. Same in the compact variant. Where timed, the new quote left 102–137 ms after the click.
- Amount too high → Apply wrote 52,000,000; the next response named a lower cap (9,800,000), since each tool states its own.
- Slippage too loose (30%, mayan / mayanMCTP / symbiosis only) → Apply set 10% and a route came back.
- Slippage too tight (eisen only, 0.001%) → Apply set 0.01%; eisen then failed with its own internal error, which is suppressed, so the generic sentence followed — the honest answer there.
- Destination account not ready (Stellar receiver without a trustline) → card, no button.
- Typing a burst sent exactly one request, 500 ms after the last key.
- One card only, in both the wide and compact variants.
- Fork: the progressive stream was checked against the live endpoint — every source emits a `routes` event, empty ones included, and the fork's stream functions build the right card from it. Exact output: the playground quoted Base → SOL without a Solana receiver, and the quote endpoint (li.quest as well) answers that with a 404 carrying no diagnostics, so the generic sentence showed; with a receiver set, the same 404 carries them (checked with curl). A failed quote shows "Quote request failed" in the panel.
- Not reached live: Apply in limit mode (limit orders have no minimum, so no card appears; unit-tested), `blockedBySettings` (needs `allowSwitchChain: false` in config), `recipientNotSupported` (captured beside a route only).

**Other notes**
- CI: `main`'s new Test workflow (#878) runs `pnpm test` on PRs, and it passes on this branch.
- Downstream: jumper-frontend never reads the routes query cache (its only `queryClient` use there is `resetQueries` on limit orders), and its `availableRoutes` handler accepts an empty list.
- AC6: the hook branch is now tested. `fromAmount: 0n` is inherent to a receive-driven quote, which has no send amount to compare; the proposal words the AC that way.

